PR #1, opened 2020-04-30
Add initial dlc functions

yancyribbens:
This PR adds simple DLC functions include CET (Contract Execution Transaction)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,0 +1,10 @@
+[package]
+name = "rust-dlc"
+version = "0.1.0"
+authors = ["Crypto Garage"]
+edition = "2018"
+
+[dependencies]
+bitcoin = "0.21"
+secp256k1 = "0.15"
+lightning = { git = "https://github.com/rust-bitcoin/rust-lightning" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,5 +6,7 @@ edition = "2018"
 
 [dependencies]
 bitcoin = "0.21"
-secp256k1 = "0.15"
-lightning = { git = "https://github.com/rust-bitcoin/rust-lightning" }
+
+[dependencies.secp256k1]
+git = "https://github.com/yancyribbens/rust-secp256k1"
+rev = "98f086355c8f05370a77a17d2c69de1115e12828"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,18 +1,149 @@
 use bitcoin::blockdata::transaction::{TxOut, TxIn, Transaction};
-use secp256k1::PublicKey;
+use bitcoin::blockdata::script::{Script, Builder};
+use bitcoin::blockdata::opcodes;
+use bitcoin::util::address::Address;
+use secp256k1::key::{PublicKey};
+use bitcoin::hashes::{Hash, sha256d};
 
-use lightning::ln::chan_utils::*;
+use secp256k1::*;
 
 const DUST_LIMIT: u64 = 5000;
 const TX_VERSION: u32 = 2;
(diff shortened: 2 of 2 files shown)


src/lib.rs
```diff
@@ -1,18 +1,146 @@
...
+
+pub fn combine_keys(pub_keys: &[PublicKey]) -> PublicKey {
+    pub_keys.iter().fold(pub_keys[0], |keys, key| keys.combine(key).unwrap())
+}
+
+pub fn get_secp_committed_key(oracle_pub_key: PublicKey,
+                              oracle_r_point: PublicKey,
+                              message: String) -> PublicKey {
+    let s = Secp256k1::signing_only();
+    let msg = Message::from_slice(message.as_bytes()).unwrap();
```
Tibo-lg, line 22: IIRC this needs to be 32 bytes so you need to hash it first.

src/lib.rs
```diff
@@ -1,18 +1,146 @@
...
+}
 
 pub fn create_funding_transaction(local_fund_pubkey: PublicKey,
-                              remote_fund_pubkey: PublicKey,
-                              output_amount: u64,
-                              local_inputs: Vec<TxIn>,
-                              remote_inputs: Vec<TxIn>,
-                              local_change_output: Vec<TxOut>,
-                              remote_change_output: Vec<TxOut>) -> Transaction {
+                                  remote_fund_pubkey: PublicKey,
```
Tibo-lg, line 138: It would be nice to use some kind of auto-formatter to avoid this kind of cosmetic changes across commits.

src/lib.rs
```diff
@@ -208,4 +359,44 @@ mod tests {
         assert_eq!(2, refund_transaction.output[1].value);
         assert_eq!(3, refund_transaction.input[0].sequence);
     }
+    #[test]
+    fn create_cet_transaction_test() {
+        let addr = Address::from_str("33iFwdLuRpW1uK1RTRqsoi8rR4NpDzk66k").unwrap();
+
+        let maturity_time = 500000001;
+        let cet_script = vec![0u8, 1, 2, 3];
```
Tibo-lg, line 367: It would be nice to also test the actual cet script creation.

Cargo.toml
```diff
@@ -1,10 +1,12 @@
 [package]
 name = "rust-dlc"
 version = "0.1.0"
-authors = ["Crypto Garage"]
+authors = ["Crypro Garage"]
```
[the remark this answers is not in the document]
yancyribbens: eeks that's a fat finger mistake.

src/lib.rs
yancyribbens: Indeed.  Vim gives me issues some times with formatting.  I should put more work into a vim setting that doesn't change.

src/lib.rs
yancyribbens: I agree.  I'll add a test for this.

src/lib.rs
yancyribbens: Thanks for catching this.  I'll update it.

yancyribbens:
@Tibo-lg I've updated per your feedback.  I'll create another repo to add to `rust-bitcoin` and make any syntax/test changes required to submit the project there.  I have not yet tested on a live version of regtest although I think this is a valuable next step.

yancyribbens:
@Tibo-lg I've updated per your feedback.  I'll create another repo to add to `rust-bitcoin` and make any syntax/test changes required to submit the project there.  I have not yet tested on a live version of regtest although I think this is a valuable next step.

Tibo-lg (approved):

PR #3, opened 2020-09-17
Feature/implement tlv messages

Tibo-lg:


dlc-messages/src/lib.rs
```diff
@@ -0,0 +1,624 @@
...
+    pub local_payout: u64,
+}
+
+impl Writeable for ContractOutcome {
+    fn write<W: Writer>(&self, writer: &mut W) -> Result<(), ::std::io::Error> {
+        let outcome = &self.outcome;
+        let nfced = outcome.nfc().collect::<String>();
+        let len = BigSize(nfced.len() as u64);
+        len.write(writer)?;
+        let bytes = nfced.as_bytes();
```
nkohen, line 52: nit: Might be nice to have factored out utility functions for serializing and deserializing `String`s? You'll need this more often in the future when you add oracle TLVs

nkohen (commented):
Looks good! Just one nit

dlc-messages/src/lib.rs
Tibo-lg: Done

PR #5, opened 2021-03-29
Feature/numerical dlcs

Tibo-lg:
Add structures and utilities to work with numeric outcome DLCs.

### DIFF
--- a/dlc/src/digit_decomposition.rs
+++ b/dlc/src/digit_decomposition.rs
@@ -1,0 +1,687 @@
+//! Utility functions to decompose numeric outcome values
+
+use super::{Error, Payout, RangePayout};
+
+/// Describes an interval that starts at `prefix || start` and terminates at `prefix || end`.
+struct PrefixInterval {
+    /// The prefix common to all numbers within the interval.
+    prefix: Vec<usize>,
+    /// The suffix of the first number in the interval.
+    start: Vec<usize>,
+    /// The suffix of the last number in the interval.
+    end: Vec<usize>,
+}
+
+/// Decompose a numeric value into digits in the specified base. If the decomposed
+/// value contains less than `nb_digits`, zeroes will be prepended to reach `nb_digits`
+/// size.
+pub fn decompose_value(mut value: usize, base: usize, nb_digits: usize) -> Vec<usize> {
+    let mut res = Vec::new();
+
+    while value > 0 {
+        res.push(value % base);
+        value = ((value as f64) / (base as f64)).floor() as usize;
+    }
+
+    while res.len() < nb_digits {
+        res.push(0);
+    }
+
+    assert_eq!(nb_digits, res.len());
+
+    res.into_iter().rev().collect()
+}
+
+/// Takes a decomposed representation of a numerical value in a given base and returns
+/// the represented value as a `usize`
+pub fn compose_value(values: &[usize], base: usize) -> usize {
+    let mut composed = 0;
+    for i in 0..values.len() {
+        let pow = values.len() - i - 1;
+        composed += values[i] * base.pow(pow as u32);
+    }
+
+    composed
+}
+
+/// Takes a vector or `RangePayout` and (if necessary) updates the first element
+/// to cover the range [0, first_end] where first_end is the end value of the
+/// first element, and updates the last element to cover the range
+/// [last_start, max_val] where last_start is the start value of the last
+/// `RangePayout` in `original_outcomes` and max_val is equal to base^nb_digits - 1.
+pub fn pad_range_payouts(
+    original_outcomes: Vec<RangePayout>,
+    base: usize,
+    nb_digits: usize,
+) -> Vec<RangePayout> {
+    let mut outcomes: Vec<RangePayout> = original_outcomes;
+    if outcomes[0].start != 0 {
+        outcomes[0] = RangePayout {
+            start: 0,
+            count: outcomes[0].count + outcomes[0].start,
+            payout: Payout {
+                offer: outcomes[0].payout.offer,
+                accept: outcomes[0].payout.accept,
+            },
+        };
+    }
+    let last_index = outcomes.len() - 1;
+    let last_outcome = &outcomes[last_index];
+    let max_value = base.pow(nb_digits as u32);
+    if last_outcome.start + last_outcome.count != max_value {
+        outcomes[last_index] = RangePayout {
+            start: last_outcome.start,
+            count: max_value - last_outcome.start,
+            payout: Payout {
+                offer: last_outcome.payout.offer,
+                accept: last_outcome.payout.accept,
+            },
+        }
+    }
+
+    outcomes
+}
+
+/// Returns the interval [start, end] as a `PrefixInterval`, which will contain
+/// the common prefix to all numbers in the interval as well as the start and end
+/// suffixes decomposed in the specified base, and zero padded to `nb_digits` if
+/// necessary.
+fn separate_prefix(
+    start: usize,
+    end: usize,
+    base: usize,
+    nb_digits: usize,
+) -> Result<PrefixInterval, Error> {
+    let start_digits = decompose_value(start, base, nb_digits);
+    let end_digits = decompose_value(end, base, nb_digits);
+    let mut prefix = Vec::new();
+
+    let mut i = 0;
+    while i < nb_digits && start_digits[i] == end_digits[i] {
+        prefix.push(start_digits[i]);
+        i += 1;
+    }
+    let start = start_digits.into_iter().skip(prefix.len()).collect();
+
+    let end = end_digits.into_iter().skip(prefix.len()).collect();
+
+    Ok(PrefixInterval { prefix, start, end })
+}
+
+/// Removes the trailing digits from `digits` that are equal to `num`.
+fn remove_tail_if_equal(mut digits: Vec<usize>, num: usize) -> Vec<usize> {
+    let mut i = digits.len();
+    while i > 1 && digits[i - 1] == num {
+        i -= 1;
+    }
+    digits.truncate(i);
+    digits
+}
+
+/// Compute the groupings for the end of the interval.
+fn back_groupings(digits: Vec<usize>, base: usize) -> Vec<Vec<usize>> {
+    let digits = remove_tail_if_equal(digits, base - 1);
+    if digits.len() == 0 {
+        return vec![vec![base - 1]];
+    }
+    let mut prefix = vec![digits[0]];
+    let mut res: Vec<Vec<usize>> = Vec::new();
+    for digit in digits.iter().skip(1) {
+        let mut last = 0;
+        let digit = digit.clone();
+        while last < digit {
+            let mut new_res = prefix.clone();
+            new_res.push(last);
+            res.push(new_res);
+            last += 1;
+        }
+        prefix.push(digit);
+    }
+    res.push(digits);
+    res
+}
+
+/// Compute the groupings for the beginning of the interval.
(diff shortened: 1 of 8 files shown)


dlc/src/digit_decomposition.rs
```diff
@@ -0,0 +1,673 @@
...
+struct PrefixRange {
+    start: Vec<usize>,
+    end: Vec<usize>,
+    prefix: Vec<usize>,
+}
+
+/// Decompose a numeric value into digits in the specified base. If the decomposed
+/// value contains less than `nb_digits`, zeroes will be prepended to reach `nb_digits`
+/// size.
+pub fn decompose_value(mut value: usize, base: usize, nb_digits: usize) -> Vec<usize> {
```
nkohen, line 14: I'm new to Rust so genuine interest question: Is it good practice to have `mut value` here in place of making a mutable copy?

dlc/src/digit_decomposition.rs
```diff
@@ -0,0 +1,673 @@
...
+    start: usize,
+    end: usize,
+    base: usize,
+    nb_digits: usize,
+) -> Result<PrefixRange, Error> {
+    let start_digits = decompose_value(start, base, nb_digits);
+    let end_digits = decompose_value(end, base, nb_digits);
+    if start_digits.len() != nb_digits || end_digits.len() != nb_digits {
+        return Err(Error::InvalidArgument);
+    }
```
nkohen, line 86: shouldn't this be a tested invariant of the `decompose_value` function? I'm not sure what the value of this check is here

dlc/src/digit_decomposition.rs
```diff
@@ -0,0 +1,673 @@
...
+    let mut prefix = Vec::new();
+
+    let mut i = 0;
+    while i < start_digits.len() && start_digits[i] == end_digits[i] {
```
nkohen, line 90: nit: `i < nb_digits` is probably more readable.

dlc/src/digit_decomposition.rs
```diff
@@ -0,0 +1,673 @@
+//! Utility functions to decompose numeric outcome values
+
+use super::{Error, Payout, RangeOutcome};
+
+struct PrefixRange {
```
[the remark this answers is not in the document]
nkohen: probably also worth having a doc describing start and end as the unique digits of the endpoints with prefix as the shared digit prefix

dlc/src/digit_decomposition.rs
```diff
@@ -0,0 +1,673 @@
...
+                offer: last_outcome.payout.offer,
+                accept: last_outcome.payout.accept,
+            },
+        }
+    }
+
+    outcomes
+}
+
+fn separate_prefix(
```
nkohen, line 76: nit: I think these functions are pretty straightforward but for someone reading this who hasn't yet looked at the spec it might be nice to have cursory doc comments on this function and the ones below

dlc/src/digit_decomposition.rs
```diff
@@ -0,0 +1,673 @@
...
+        i += 1;
+    }
+    let start = start_digits.into_iter().skip(prefix.len()).collect();
+
+    let end = end_digits.into_iter().skip(prefix.len()).collect();
+
+    Ok(PrefixRange { prefix, start, end })
+}
+
+fn remove_back_if_equal(mut digits: Vec<usize>, num: usize) -> Vec<usize> {
```
nkohen, line 101: another Rust question: is it usual to both mutate the input and return it? I know that in C you usually do one or the other though sometimes in Java you also return if you want a "fluent interface."

nkohen: Sorry for all the noob questions

nkohen, line 101: nit: one of `remove_suffix_if_equal` or `remove_tail_if_equal` or `truncate_tail_if_equal` might be slightly more descriptive as "back" is a bit vague imo

dlc/src/lib.rs
```diff
@@ -65,14 +68,36 @@ const ENABLE_LOCKTIME: u32 = 0xfffffffe;
...
 pub struct Payout {
     /// Payout for the offering party
     pub offer: u64,
     /// Payout for the accepting party
     pub accept: u64,
 }
 
+#[derive(PartialEq, Debug)]
+/// Representation of a set of contiguous outcomes that share a single payout.
+pub struct RangeOutcome {
```
nkohen, line 81: nit: I'm not sure this is the most descriptive name, in bitcoin-s we use `ConstantPayoutRange` I think

dlc/src/lib.rs
```diff
@@ -65,14 +68,36 @@ const ENABLE_LOCKTIME: u32 = 0xfffffffe;
...
+    /// The start of the range
+    pub start: usize,
+    /// The number of outcomes in the range
+    pub count: usize,
+    /// The payout associated with all outcomes
+    pub payout: Payout,
+}
+
+#[derive(PartialEq, Debug)]
+/// Structure that stores which stores CET and adaptor signature information.
```
nkohen, line 91: nit: "that stores which stores"

dlc/src/lib.rs
```diff
@@ -65,14 +68,36 @@ const ENABLE_LOCKTIME: u32 = 0xfffffffe;
...
+pub struct RangeInfo {
+    /// a cet index
+    pub cet_index: usize,
+    /// an adaptor signature index
+    pub adaptor_index: usize,
```
nkohen, line 96: At first glance I'm not actually sure I know what these indices represent?

dlc/src/lib.rs
```diff
@@ -417,7 +451,53 @@ pub fn make_funding_redeemscript(a: &PublicKey, b: &PublicKey) -> Script {
         .into_script()
 }
 
-/// Create an adaptor signature for the given cet
+fn combine_pubkeys(pubkeys: &Vec<PublicKey>) -> Result<PublicKey, Error> {
```
nkohen, line 454: isn't there a function on secp256k1's public API which takes in an array of pub keys?

dlc/src/digit_trie.rs
```diff
@@ -0,0 +1,637 @@
...
+    /// satisfy the borrow checker.  
+    store: Vec<Node<DigitLeaf<T>, DigitNode<T>>>,
+    root: Option<usize>,
+    base: usize,
+}
+
+/// Structure used to iterated through a `DigitTrie` values.
+pub struct DigitTrieIter<'a, T> {
+    trie: &'a DigitTrie<T>,
+    index_stack: Vec<(Option<usize>, isize)>,
```
nkohen, line 18: doc comment for `index_stack` would be helpful

dlc/src/digit_decomposition.rs
```diff
@@ -0,0 +1,673 @@
...
+        let pow = values.len() - i - 1;
+        composed += values[i] * base.pow(pow as u32);
+    }
+
+    composed
+}
+
+/// Takes a vector or `RangeOutcome` and (if necessary) adds one at the beginning
+/// and at the end of it to cover the full range of possible outcome values
+pub fn get_max_ranges(
```
nkohen, line 43: perhaps a more descriptive name would be `pad_...` this name seems really misleading, it took me a while to figure out what this function was doing. (also the doc description of "adds one at the beginning and at the end" didn't mean much because I had to read all the code to figure out the answer to "one of what" and also it isn't adding, it is modifying the extremal ranges to extend to the boundaries).

nkohen (commented):
Looking pretty good :)

FYI, I didn't look too closely at tests or the trie implementation

Things that should have types to make the code a little bit more readable:
`Vec<usize>` (e.g. `Digits`, or potentially multiple types if you want to make it clearer when you have full digits or just a prefix and things like this)
`Vec<Vec<usize>>` (i.e. Vec of one of the above types)
`Vec<Message>` (e.g. `Outcome`)

dlc/src/digit_decomposition.rs
Tibo-lg: Because `usize` is a primitive type it is actually copied when passed by value (if that makes sense). The `mut` here just indicates that the variable `value` can be mutated within the function. I don't know if it's good practice or not, but since it's copied I don't see what could be bad about it.

dlc/src/digit_decomposition.rs
Tibo-lg: > Sorry for all the noob questions

No problem, please ask them because I'm definitely not an expert and it forces me to double check things :)

> is it usual to both mutate the input and return it?

So in that case `digits` is passed by value, which means that ownership of it is transferred to the current function, so after calling this function, the variable passed as first parameter cannot be used anymore in the calling function. Which is why it needs to return it as well otherwise the calling function would have just "lost it".

I think in that case it is possible to pass is as a mutable reference as well, in which case it wouldn't need to be returned. I don't recall exactly why I did it like that but I don't think it's especially wrong. If you think it'd be clearer using a mutable reference I don't mind updating.

dlc/src/lib.rs
Tibo-lg: Not that I'm aware off, and checked quickly but couldn't find. Happy to hear if there is one though.

dlc/src/digit_decomposition.rs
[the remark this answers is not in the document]
Tibo-lg: Renamed and added doc, hope it's clearer.

dlc/src/digit_decomposition.rs
Tibo-lg: You're right, removed here and added an `assert` in `decompose_value`

dlc/src/digit_decomposition.rs
Tibo-lg: done

dlc/src/digit_decomposition.rs
Tibo-lg: Added some doc hope it helps

dlc/src/digit_decomposition.rs
Tibo-lg: good point, went with `remove_tail_if_equal`

dlc/src/lib.rs
Tibo-lg: Went with `RangePayout`, `ConstantPayoutRange` felt a bit long.

dlc/src/lib.rs
Tibo-lg: fixed thanks

dlc/src/lib.rs
Tibo-lg: Tried to be a bit more descriptive, but basically the point is that CETs and adaptor signatures are kept into vectors, and so these fields store the index in those. It's a bit difficult to understand without context I must agree, but not sure how to make it clearer.

dlc/src/digit_trie.rs
Tibo-lg: Added as well as a bit more info about the type of traversal performed by the iterator.

dlc/src/digit_decomposition.rs
Tibo-lg: Good point. Went with `pad_range_payouts` and tried to write a better doc for it. Also the doc was definitely wrong about the adding part thanks for spotting!

dlc/src/lib.rs
nkohen: https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1.h#L760

Tibo-lg: Oh thanks I missed that! Looks like in rust-secp256k1 they only use it to combine two pubkeys for now though: https://github.com/rust-bitcoin/rust-secp256k1/blob/master/src/key.rs#L369

I'll try to do a PR to add a version with multiple keys but it might take a bit time before it gets merged.

PR #66, opened 2022-06-24
Added a pre-closed state to contract in manager

jatayl:
Added a pre-closed state to contract in manager that will fully close when the cet is confirmed to the blockchain


dlc-manager/src/manager.rs
```diff
@@ -903,6 +905,45 @@ where
...
+        Ok(())
+    }
+
+    fn check_preclosed_contract(&mut self, contract: &PreClosedContract) -> Result<(), Error> {
+        let broadcasted_cet = contract
+            .signed_contract
+            .accepted_contract
+            .dlc_transactions
+            .cets[contract.cet_index]
+            .clone();
```
Tibo-lg, line 928: nit: I think you could just work with a reference without cloning in that case.

dlc-manager/src/manager.rs
```diff
@@ -903,6 +905,45 @@ where
...
+        let broadcasted_txid = broadcasted_cet.txid();
+        let confirmations = self
+            .wallet
+            .get_transaction_confirmations(&broadcasted_txid)
+            .unwrap();
```
Tibo-lg, line 933: I think it'd be better to return an error than unwrap (just `?`).

dlc-manager/src/manager.rs
```diff
@@ -903,6 +905,45 @@ where
...
+        if confirmations >= 1 {
```
Tibo-lg, line 934: Could you use `NB_CONFIRMATIONS` instead of `1`? 6 is usually the number considered safe.

dlc-manager/src/contract/mod.rs
```diff
@@ -149,8 +160,19 @@ pub struct FailedSignContract {
     pub error_message: String,
 }
 
+/// Information about a contract that is almost closed by a broadcasted, but not confirmed CET.
+#[derive(Clone)]
+pub struct PreClosedContract {
```
Tibo-lg, line 165: Since this struct is exactly the same as `ClosedContract` it might be ok to just use the same struct but a different `Contract` enum variant like is already done for e.g. `Signed` and `Confirmed`.

dlc-manager/src/manager.rs
```diff
@@ -903,6 +905,45 @@ where
         Ok(())
     }
 
+    fn check_preclosed_contracts(&mut self) -> Result<(), Error> {
+        for c in self.store.get_preclosed_contracts()? {
+            if let Err(e) = self.check_preclosed_contract(&c) {
+                error!(
+                    "Error checking confirmed contract {}: {}",
```
Tibo-lg, line 912: nit: ~confirmed~ pre-closed

dlc-manager/src/manager.rs
```diff
@@ -976,16 +1017,25 @@ where
...
+                    signed_contract: contract.clone(),
+                    attestations: attestations.iter().map(|x| x.1.clone()).collect(),
+                    cet_index: range_info.cet_index,
+                };
 
-            self.store
-                .update_contract(&Contract::Closed(closed_contract))?;
+                self.store
+                    .update_contract(&Contract::PreClosed(preclosed_contract))?;
+            } else {
```
Tibo-lg, line 1029: I think the ideal would be to have a case for `1 <= confirmations < 6` in which it's pre-closed, and then this else.

jatayl:
I will update the integration tests with the new contract state flow

Tibo-lg (changes requested):
Thanks a lot for the PR! Mainly minor comments. You'll also need to update the integration tests for the CI to pass. Ideally it'd be nice to have the test check for the pre-close state as well, and also some unit tests for `get_preclosed_contracts` in `dlc-sled-storage-provider` but I can add those after.

dlc-manager/src/contract/mod.rs
jatayl: Makes sense. I left it as two individual structs with the intention of cutting down on the number of fields in the `ClosedContract` struct as a means of pruning in the future. Would you still prefer one shared over two individual? I agree that if no pruning of information in the `ClosedContract` struct is to occur, they should share.

jatayl:
@Tibo-lg appreciate the review! I'll try to address the unit and integration tests in this PR

dlc-manager/src/manager.rs
jatayl: exposed by copy and paste :)

Tibo-lg: Yeah I've made the same line 832...

dlc-manager/src/contract/mod.rs
Tibo-lg: Mmmm fair enough, maybe ok to keep it as is then, would you maybe just add a todo or comment?

jatayl: Sure!

Tibo-lg (approved):

PR #67, opened 2022-07-03
Update Development.md file; fix test command

sosaucily:
Add missing -- and add info about running all tests.

### DIFF
--- a/docs/Development.md
+++ b/docs/Development.md
@@ -21,7 +21,7 @@ For example within the `dlc-manager` folder:
 cargo test -- --ignored two_of_five_oracle_numerical_test
 ```
 
-Or run all the integration tests with 
+Or run all the tests including all integration tests with 
 ```
 cargo test -- --include-ignored
 ```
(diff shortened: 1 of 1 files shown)


docs/Development.md
```diff
@@ -13,12 +13,17 @@ In the `dlc` or `dlc-manager` directory, run `../scripts/start_node.sh` to start
...
 replacing `test_name` with the test you want to run.
 For example within the `dlc-manager` folder:
 ```
-cargo test --ignored two_of_five_oracle_numerical_test
+cargo test -- --ignored two_of_five_oracle_numerical_test
+```
+
+Or run all the integration tests with 
+```
+cargo test -- --include-ignored
```
Tibo-lg, line 26: This will run not only the integration tests but also the unit tests, so might be nice to say something like `Or run all the test including integration tests with`.

Tibo-lg (changes requested):
Thanks for the fix, tiny nit

PR #69, opened 2022-07-04
Update example readme for clarity

sosaucily:


Tibo-lg:
There was a bug that made it display `Rejected` instead of the actual state, fixed in #70. With the fix I don't thing further clarification is required as the readme already state that you should generate blocks after the transaction is broadcast.

Thanks for spotting the issue though!

PR #71, opened 2022-07-05
Prune closed contract data

### LINKED ISSUE #61: Prune contract state once closed
At the moment all contract information is kept even after the contract has been closed. It would be better to prune it as to reduce storage size.

jatayl:
This PR attempts to fix #61


dlc-manager/src/contract/accepted_contract.rs
```diff
@@ -50,6 +50,31 @@ impl AcceptedContract {
...
+            .find_map(|x| {
+                if &x.script_pubkey == v0_witness_payout_script {
+                    Some(x.value)
+                } else {
+                    None
+                }
+            })
+            .unwrap_or(0) as i64;
+        final_payout - collateral
+    }
```
jatayl, line 77: I moved this implementation from the sample code into a method for accepted contracts, so that pnl can be computed internally. Perhaps this means unit tests should be written for this method?

dlc-manager/src/manager.rs
```diff
@@ -925,15 +925,27 @@ where
...
+                    .id,
+                counter_party_id: contract
+                    .signed_contract
+                    .accepted_contract
+                    .offered_contract
+                    .counter_party,
+                pnl: contract
+                    .signed_contract
+                    .accepted_contract
+                    .compute_pnl(contract.cet_index),
```
jatayl, line 948: A big verbose here. curious of your thoughts

dlc-messages/src/ser_impls.rs
```diff
@@ -537,6 +538,23 @@ pub fn read_i32<R: ::std::io::Read>(reader: &mut R) -> Result<i32, DecodeError>
...
+    i2.write(writer)
+}
+
+/// Reads an `i64` value from the given reader.
+pub fn read_i64<R: ::std::io::Read>(reader: &mut R) -> Result<i64, DecodeError> {
+    let v1: [u8; 4] = Readable::read(reader)?;
+    let v2: [u8; 4] = Readable::read(reader)?;
+    let v = [v1, v2].concat();
+    Ok(i64::from_be_bytes(v.try_into().unwrap()))
+}
```
jatayl, line 556: Because pnl is i64, i64 needs to be readable / writable. I tried to mirror your implementation for i32, but it seems that Readable/Writeable in rust-lightning are not implemented for [u8; 8].  im not a huge fan of line 554 because i believe concat creates a heap allocation by going through a Vec first.

with that being said, on the nightly compiler, the concat method can simply generate an array without going through vec.

https://doc.rust-lang.org/std/slice/trait.Concat.html

Tibo-lg, line 556: What about just writing/reading byte per byte? Did some quick benchmark and seems much faster for the reading (somehow the writing benchmark doesn't work so couldn't compare).
```suggestion
/// Writes an `i64` value to the given writer.
pub fn write_i64<W: Writer>(i: &i64, writer: &mut W) -> Result<(), ::std::io::Error> {
    let i = i.to_be_bytes();
    for b in i {
        b.write(writer)?;
    }
    Ok(())
}

/// Reads an `i64` value from the given reader.
pub fn read_i64<R: ::std::io::Read>(reader: &mut R) -> Result<i64, DecodeError> {
    let mut v = [0u8; 8];
    for i in 0..8 {
        v[i] = Readable::read(reader)?;
    }
    Ok(i64::from_be_bytes(v))
}
```

dlc-manager/src/manager.rs
Tibo-lg: Yeah I don't think you can do better. Tbh I'm not super happy with the way I structured contract. The goal was to avoid copying fields between structs and working around the lack of inheritance, but I think copying field might be better at the end (which is the route I took for the channels)

dlc-manager/src/contract/accepted_contract.rs
Tibo-lg: Unit tests would be great, one thing that might be annoying is constructing the contract, maybe you can use some json input or something to make it easier.

Tibo-lg (changes requested):
Thanks! Looks good, just the serialization part to update and if you're motivated the unit test for the pnl computation.

dlc-messages/src/ser_impls.rs
jatayl: that crossed my mind, and after seeing it implemented, i like it a lot more! lets go with it

jatayl:
thanks for the review. i liked and accepted de/serializing byte by byte. i will get to the unit tests

dlc-messages/src/ser_impls.rs
Tibo-lg: Wooops made clippy unhappy sorry...

Tibo-lg:
@jatayl I hope you don't mind I'll finish this.

jatayl:
@Tibo-lg sounds good! Sorry I got wrapped up in other work and completely forgot about the PR. Thanks for handling it.

PR #97, opened 2023-03-01
Feature/ln dlc channels

Tibo-lg:


luckysori:
Oops, force-pushed here by mistake. Sorry!

I've now corrected it but should we protect this branch just in case, @Tibo-lg?

Tibo-lg:
> Oops, force-pushed here by mistake. Sorry!
> 
> I've now corrected it but should we protect this branch just in case, @Tibo-lg?

Good idea, done!

PR #103, opened 2023-04-17
Update `std::error::Error` implementations

luckysori:
My main motivation is to make `rust-dlc` work better with `anyhow`, but I cleaned up some minor things along the way.


Tibo-lg (approved):
Thanks!

PR #105, opened 2023-04-20
Add more trait bounds to `dlc_manager::Error::WalletError`

luckysori:
This ensures that the `dlc_manager::Error` enum can be used with `anyhow`.


Tibo-lg (approved):
Mmm ok I guess it's fine.

PR #107, opened 2023-04-21
Feature/ln dlc channels debug

da-kami:
The path where the multiple DLCs are being created was not actually called in the test `ln_dlc_off_chain_close_open_close`.

Once we call them the test fails with:

```
...


From Bob: 2023-04-21 20:24:12.843 DEBUG [lightning::chain::chainmonitor:708] Persistence of ChannelMonitorUpdate for channel c7346e546f15e1af92bf42305bbfaa8b287c32871954b448854eb1df785f59ca completed

thread 'ln_dlc_off_chain_close_open_close' panicked at 'called `Result::unwrap()` on an `Err` value: InvalidParameters("Invalid split revocation secret")', dlc-manager/tests/ln_dlc_channel_execution_tests.rs:1669:10
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
test ln_dlc_off_chain_close_open_close ... FAILED

failures:

failures:
    ln_dlc_off_chain_close_open_close

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 11 filtered out; finished in 17.65s
```

### DIFF
--- a/dlc-manager/tests/ln_dlc_channel_execution_tests.rs
+++ b/dlc-manager/tests/ln_dlc_channel_execution_tests.rs
@@ -926,7 +926,7 @@ fn ln_dlc_test(test_path: TestPath) {
             return;
         }
 
-        off_chain_close_finalize(&test_params, &alice_node, &bob_node, channel_id);
+        off_chain_close_finalize(&alice_node, &bob_node, channel_id);
 
         if let TestPath::OffChainCloseOpenClose = test_path {
             offer_sub_channel(
@@ -939,7 +939,7 @@ fn ln_dlc_test(test_path: TestPath) {
                 bob_descriptor.clone(),
             );
             off_chain_close_offer(&test_params, &alice_node, &bob_node, channel_id);
-            off_chain_close_finalize(&test_params, &alice_node, &bob_node, channel_id);
+            off_chain_close_finalize(&alice_node, &bob_node, channel_id);
         }
 
         offer_sub_channel(
@@ -1444,12 +1444,7 @@ fn offer_sub_channel(
     assert_sub_channel_state!(alice_node.sub_channel_manager, channel_id, Confirmed);
 
     if let TestPath::Reconnect = test_path {
-        reconnect(
-            alice_node,
-            bob_node,
-            alice_descriptor.clone(),
-            bob_descriptor.clone(),
-        );
+        reconnect(alice_node, bob_node, alice_descriptor, bob_descriptor);
 
         // For some weird reason uncommenting this triggers a stack overflow...
         // assert_sub_channel_state!(alice_node.sub_channel_manager, channel_id, Confirmed);
@@ -1580,8 +1575,7 @@ fn off_chain_close_offer(
         .unwrap();
 
     let dlc_channel_id = sub_channel.get_dlc_channel_id(0).unwrap();
-    let contract_id =
-        assert_channel_contract_state!(alice_node.dlc_manager, dlc_channel_id, Confirmed);
+    assert_channel_contract_state!(alice_node.dlc_manager, dlc_channel_id, Confirmed);
 
     let (close_offer, _) = alice_node
         .sub_channel_manager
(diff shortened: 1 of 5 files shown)


luckysori:
@Tibo-lg: With https://github.com/p2pderivatives/rust-dlc/pull/107/commits/d7f0a747082b7ca3f69b400ee7795c2548bebb0d it looks like the bug is fixed now. The test passes.

luckysori:
I've added https://github.com/p2pderivatives/rust-dlc/pull/107/commits/37d243891d86b98e741772d6b68a60e97311b065 since even your latest patch on `ln-dlc-channels` failed.

luckysori:
> I've added [37d2438](https://github.com/p2pderivatives/rust-dlc/commit/37d243891d86b98e741772d6b68a60e97311b065) since even your latest patch on `ln-dlc-channels` failed.

Actually, that was just a symptom. It's still failing, probably because there are 2 tests sharing resources. Maybe we just need to run them sequentially.

dlc-manager/tests/ln_dlc_channel_execution_tests.rs
```diff
@@ -935,7 +939,21 @@ fn ln_dlc_test(test_path: TestPath) {
                 bob_descriptor.clone(),
             );
             off_chain_close_offer(&test_params, &alice_node, &bob_node, channel_id);
-            off_chain_close_finalize(&test_params, &alice_node, &bob_node, channel_id);
+            off_chain_close_finalize(&alice_node, &bob_node, channel_id);
+        }
+
+        if let TestPath::OffChainCloseOpenClose = test_path {
```
Tibo-lg, line 945: Wasn't it enough to add an open/close? Are two required? Also I don't think you need another `if let` here?

luckysori: > Wasn't it enough to add an open/close? Are two required?

I could only reproduce the bug by opening, closing, opening and closing a second time. I think it's because the error path can only be hit if there was at least one other secret generated before:

```rust
/// Inserts the `secret` at `idx`. Returns `Ok(())` if the secret
/// was generated in accordance with BOLT 3 and is consistent with previous secrets.
pub fn provide_secret(&mut self, idx: u64, secret: [u8; 32]) -> Result<(), ()> {
    let pos = Self::place_secret(idx);
    // If we place the secret in position 0, the for loop is skipped so there is no error path
    for i in 0..pos {
        let (old_secret, old_idx) = self.old_secrets[i as usize];
        if Self::derive_secret(secret, pos, old_idx) != old_secret {
            return Err(());
        }
    }
    if self.get_min_seen_secret() <= idx {
        return Ok(());
    }
    self.old_secrets[pos as usize] = (secret, idx);
    Ok(())
}
```

> Also I don't think you need another if let here?

Do you mean that this would be the third open-close? I agree, I think there's one cycle too many.

Tibo-lg: > Do you mean that this would be the third open-close? I agree, I think there's one cycle too many.

Yes! And even if there was a need for an extra cycle, the `if` is redundant AFAICT (it's already within an if block with the same condition).

luckysori: That is true (I didn't write this code 😛).

dlc-manager/tests/ln_dlc_channel_execution_tests.rs
```diff
@@ -1568,17 +1567,6 @@ fn off_chain_close_offer(
...
-        .dlc_manager
-        .get_store()
-        .get_sub_channel(channel_id)
-        .unwrap()
-        .unwrap();
-
-    let dlc_channel_id = sub_channel.get_dlc_channel_id(0).unwrap();
-    let contract_id =
-        assert_channel_contract_state!(alice_node.dlc_manager, dlc_channel_id, Confirmed);
-
```
Tibo-lg, line 1581: Is there a reason to remove this? I checked and the test still passes with it so I'd rather keep it.

Tibo-lg: I can see the unused `contract_id` generated a warning, but would be better to keep the assert.

luckysori: No reason to remove it other than Clippy. I probably made a mistake. I'll fix it.

luckysori: Yeah, I just deleted it too much D:

Tibo-lg (approved):
Thanks!

PR #108, opened 2023-04-22
Add global-context flag on secp256k1 on mocks

sosaucily:
global-context flag is needed on the secp256k1 package to compile this mocks package on it's own. Was inheriting from other imports when compiling the whole repo


Tibo-lg (approved):
Thanks!

PR #109, opened 2023-04-24
Flush sled tree on upsert channel and sub_channel

luckysori:
We needed this as otherwise we'd run into invalid states after restarting the LDK node.


luckysori:
Wrong branch, oops.

PR #110, opened 2023-04-24
Flush sled tree on upsert channel and sub_channel

luckysori:
Otherwise we can run into invalid-state errors when the node is restarted.

### DIFF
--- a/dlc-sled-storage-provider/src/lib.rs
+++ b/dlc-sled-storage-provider/src/lib.rs
@@ -361,6 +361,8 @@ impl Storage for SledStorageProvider {
                 },
             )
         .map_err(to_storage_error)?;
+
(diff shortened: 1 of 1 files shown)


dlc-sled-storage-provider/src/lib.rs
```diff
@@ -361,6 +361,8 @@ impl Storage for SledStorageProvider {
                 },
             )
         .map_err(to_storage_error)?;
+
+        channel_tree.flush().map_err(|e| Error::StorageError(e.to_string()))?;
```
Tibo-lg, line 365: nit: can you use `to_storage_error` (just to be consistent with the rest of the code)

Tibo-lg (approved):

PR #115, opened 2023-06-24
Update get_utxos fn, fix locked_utxos bug

sosaucily:
I think you meant for this logic.


Tibo-lg (approved):
Yes that makes sense thanks!

PR #121, opened 2023-07-10
Feat/periodic check returns affected contracts

sosaucily:


dlc-manager/src/lib.rs
```diff
@@ -199,3 +199,14 @@ impl_dlc_writeable!(Utxo, {
     (redeem_script, writeable),
     (reserved, writeable)
 });
+
+/// The response from periodic_check fn which returns the ids of the contracts that were affected in each state.
+#[derive(Clone, Debug)]
+pub struct AffectedContractIDs {
```
Tibo-lg, line 205: nit: can we call that `UpdatedContractIds` maybe?

dlc-manager/src/manager.rs
```diff
@@ -470,47 +472,59 @@ where
         Err(e)
     }
 
-    fn check_signed_contract(&mut self, contract: &SignedContract) -> Result<(), Error> {
+    fn check_signed_contract(
+        &mut self,
+        contract: &SignedContract,
+    ) -> Result<Option<ContractId>, Error> {
```
Tibo-lg, line 478: It feels like it'd be easier to just return a boolean here, indicating whether the contract was updated or not.

dlc-manager/src/manager.rs
```diff
@@ -2267,4 +2293,29 @@ mod test {
             .on_dlc_message(&offer_message, pubkey())
             .expect_err("To reject the second offer message");
     }
+
+    #[test]
```
Tibo-lg, line 2297: that's not a great test, I think it'd be better to add some checks in the integrations tests.

dlc/Cargo.toml
```diff
@@ -9,7 +9,7 @@ version = "0.4.0"
 
 [dependencies]
 bitcoin = {version = "0.29.2"}
-miniscript = "8.0.0"
+miniscript = { version = "9.0", features = ["serde", "std"], default-features = false }
```
Tibo-lg, line 12: What is the reason for this change? Also it'd be better to put the serde feature behind the `use-serde` feature flag.

simple-wallet/Cargo.toml
```diff
@@ -10,7 +10,7 @@ bitcoin = "0.29"
 dlc = {path = "../dlc"}
 dlc-manager = {path = "../dlc-manager"}
 lightning = {version = "0.0.113"}
-bdk = {version = "0.28.0"}
+bdk = {version = "0.28.0", default-features = false, features = ["async-interface", "use-esplora-async"]}
```
Tibo-lg, line 13: It doesn't feel like this is required for it to work?

Tibo-lg (changes requested):
Concept ack. Just feel like it could be made more simple with a boolean instead of an option. Also unsure what are the changes to the cargo files.

dlc-manager/src/manager.rs
sosaucily: smart!!!

PR #122, opened 2023-07-11
Split block processing from confirmation reporting

luckysori:


dlc-manager/src/chain_monitor.rs
```diff
@@ -108,41 +107,91 @@ impl ChainMonitor {
...
-                    let info = self.watched_tx.get(txid);
-                    if info.is_some() {
-                        return info;
-                    }
-                }
-                None
-            });
-            if let Some(channel_info) = channel_info {
-                res.push((tx.clone(), channel_info.clone()));
+        for (txid, state) in self.watched_tx.iter_mut() {
```
Tibo-lg, line 118: Since `watched_tx` is a hash map, it feels like it would be more efficient to iterate the tx in the block and check if they are watched ones. Now it will most of the time iterate all block transaction many times.

dlc-manager/src/chain_monitor.rs
```diff
@@ -108,41 +107,91 @@ impl ChainMonitor {
...
+                *self = WatchState::Confirmed {
+                    channel_info: *channel_info,
+                    transaction,
+                }
+            }
+            WatchState::Confirmed {
+                channel_info,
+                transaction,
+            } => {
+                log::warn!(
```
Tibo-lg, line 183: nit: i'd probably put that at the error level unless there is some cases where this should happen without it being an error?

dlc-manager/src/chain_monitor.rs
```diff
@@ -108,41 +107,91 @@ impl ChainMonitor {
...
+    ///
+    /// # Panics
+    ///
+    /// Panics if the new block's height is not exactly one more than the last processed height.
+    pub(crate) fn process_block(&mut self, block: &Block, height: u64) {
         assert_eq!(self.last_height + 1, height);
 
-        for tx in &block.txdata {
-            let channel_info = self.watched_tx.get(&tx.txid()).or_else(|| {
-                for txid in tx.input.iter().map(|x| &x.previous_output.txid) {
```
Tibo-lg, line 122: Didn't the `or_else` part disappear in your version?

Tibo-lg (commented):
Concept ack. Some logic might have been lost in the change, apart from that just nits. I would probably need to review it at least one more time though.

dlc-manager/src/chain_monitor.rs
luckysori: I think you are totally right that it makes way more sense to iterate through the block to take advantage of the constant-time lookup in the hashmap.

I guess it does depend on the size of the hashmap compared to the size of the block, but I don't think we need to worry about that now. I'll go back to the original!

dlc-manager/src/chain_monitor.rs
luckysori: You are right! And it wasn't on purpose.

But looking at it now, what is the purpose of it? It seems like for every transaction in the block we:

1. Look it up in our hashmap.
2. If not in the hashmap, we look up every input to the transaction in the hashmap too.

Is the second step to ensure that we don't miss transactions? Given the assertion above, I don't think we can miss blocks without triggering the panic. Am I missing something?

dlc-manager/src/chain_monitor.rs
luckysori: Fair enough, will do!

dlc-manager/src/chain_monitor.rs
Tibo-lg: To be honest it might not be the prettiest, but the point is to detect both transactions that are registered, and transactions that spend those. It's required by the logic of the channel manager, to detect when a buffer transaction is spent, as it would be a pain to register all CETs. Feel free to refactor if you think of something better :)

luckysori: Oh, interesting! I hadn't considered that. It's obviously a bit surprising so I would just try to document it clearly for now.

In terms of a possible rework, I guess what we want is the ability to watch transactions _and_ transaction outputs. `rust-lightning` offers the same two options with the `chain::Filter` trait.

---

Btw, I thought about reusing `rust-lightning` for this stuff but I think it's not really possible, because those traits are built to be implemented by consumers but used by LDK, which makes sense. It would be pretty weird to plug into that, but we could copy the structure so that `ChainMonitor` can be replaced with a trait.

Tibo-lg: > Oh, interesting! I hadn't considered that. It's obviously a bit surprising so I would just try to document it clearly for now.

Yeah definitely not super clear, sorry :)

> In terms of a possible rework, I guess what we want is the ability to watch transactions and transaction outputs. rust-lightning offers the same two options with the chain::Filter trait.

Yeah I guess that's it!

> but we could copy the structure so that ChainMonitor can be replaced with a trait.

Yeah that's definitely what I'd like to do!

luckysori:
Let me know what you think about https://github.com/p2pderivatives/rust-dlc/pull/122/commits/cabf2feba5282d815f4abc8103f78dc6f79f4570, @Tibo-lg. I've tried to address your comments.

I made it so that we model watching the buffer transaction output a bit more explicitly. I think it makes the code easier to follow. Since I did that it had the effect of being able to use the `TxType::Cet` to also untangle the code which handles a confirmed CET too.

Tibo-lg (approved):
I think it's nice as it makes the code clearer, though I kind of would have preferred if the `periodic_chain_monitor` method was completely removed from both managers, but I guess it can be done later.

luckysori:
I'm gonna wait for #128 before merging this because there will be conflicts and the other PR is way more important.

dlc-manager/src/chain_monitor.rs
```diff
@@ -79,61 +82,145 @@ impl ChainMonitor {
...
+        // the buffer transaction _output_ so that we can detect when
+        // a CET spends it without having to watch every possible CET
+        if channel_info.tx_type == TxType::BufferTx {
+            let outpoint = OutPoint {
+                txid,
+                // We can safely assume that the buffer transaction
+                // only has one output
+                vout: 0,
+            };
+            self.add_txo(
```
Tibo-lg, line 98: I wonder if it would be better to watch the output only once we see the transaction itself.

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -2840,282 +2840,255 @@ where
         msgs
     }
 
-    /// Updtates the view of the blockchain processing transactions and acting upon them if
+    /// Updates the view of the blockchain processing transactions and acting upon them if
```
Tibo-lg, line 2843: Maybe might as well update the comment, IIUC this is not updating the view of the blockchain anymore since it's done separately?

Tibo-lg (commented):
Git really didn't do a good job, the diff is really hard to review. But I think apart from the small comments that I left it looks good!

luckysori:
> Git really didn't do a good job, the diff is really hard to review. But I think apart from the small comments that I left it looks good!

I think when it comes to rebases it can be a bit of a disaster. Sorry about that!

dlc-manager/src/chain_monitor.rs
luckysori: It would be a bit more efficient, because now we are doing a bit of unnecessary work for CETs that definitely cannot be there yet.

But the implementation is very simple and the workload added is minimal: 
 - One TXO for all CETs.
 - No overhead when processing a new block because we are looking up in the `watched_txo` map per txin of each transaction _in the block_ anyway.
 - Small overhead when calling `confirmed_txs` as we have to filter out all the txos that could not possibly be confirmed yet.

dlc-manager/src/sub_channel_manager.rs
luckysori: Good catch, I clearly didn't read what I updated lol.

PR #123, opened 2023-07-17
Clear certain unprocessed messages after peer disconnect

luckysori:
The DLC message handler implements the `OnionMessageHandler` and `OnionMessageProvider` to be able to hook onto the `peer_disconnected` event. This allows for clearing certain `rust-dlc` messages precisely when a disconnect happens.

We drop the `SubChannelConfirm` and `SubChannelCloseConfirm` messages to avoid bugs such as the one reported [here](https://github.com/get10101/10101/issues/792).

The rest of the implementations of `OnionMessageHandler` and `OnionMessageProvider` is copied from the `IgnoringMessageHandler`.


Tibo-lg (approved):
LGTM

holzeis (approved):
👍

PR #124, opened 2023-07-17
Minor improvements

luckysori:
Sorry for combining these into a single PR, but it's a bit easier like this and I hope they are not controversial.

### DIFF
--- a/dlc-manager/src/error.rs
+++ b/dlc-manager/src/error.rs
@@ -16,7 +16,7 @@ pub enum Error {
     /// An invalid state was encounter, likely to indicate a bug.
     InvalidState(String),
     /// An error occurred in the wallet component.
-    WalletError(Box<dyn std::error::Error>),
+    WalletError(Box<dyn std::error::Error + Send + Sync + 'static>),
     /// An error occurred in the blockchain component.
     BlockchainError(String),
     /// The storage component encountered an error.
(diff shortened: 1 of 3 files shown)


dlc-manager/src/error.rs
```diff
@@ -16,7 +16,7 @@ pub enum Error {
     /// An invalid state was encounter, likely to indicate a bug.
     InvalidState(String),
     /// An error occurred in the wallet component.
-    WalletError(Box<dyn std::error::Error>),
+    WalletError(Box<dyn std::error::Error + Send + Sync + 'static>),
```
Tibo-lg, line 19: I'm a bit confused why you need that as it's not required by the library?

luckysori: I tried to explain it in the commit message:

> This ensures that the `dlc_manager::Error` enum can be used with `anyhow`.

These bounds are required by `anyhow`. Since the enum variant is generic we need to add the bounds if we want to stay compatible.

My motivation was to be able to do `.context()` or simply `?` on this error type into `anyhow::Result` in our `10101` project. Since `anyhow` is the de facto way to do error handling in Rust applications I think it's generally a good idea.

Tibo-lg: Ah somehow didn't see the full commit message. I guess it's fine.

Tibo-lg (approved):

PR #126, opened 2023-07-19
Log deser errors when getting all channels/contracts

luckysori:
Before this patch we were immediately failing the entire call to `get_contracts` or `get_sub_channels` in the event of encountering a single deserialisation failure. This can be problematic if we introduce breaking changes and consumers keep around incompatible data.

---

We absolutely don't need to merge this if you think it's too much, @Tibo-lg. In practice, this is a mandatory patch for us at 10101. It's kind of nice to be able to be resilient against these things and it's probably okay since `dlc-sled-storage-provider` is kind of a utility crate.


Tibo-lg (approved):
Seems reasonable to me, and quite a small change so no problem.

PR #131, opened 2023-07-27
Handle new `Revoke` message in `CustomMessageHandler` impl

luckysori:
I was surprised to see that this wasn't caught by the tests 😮 We only discovered this when upgrading the `rust-dlc` dependency for the 10101 repo.


Tibo-lg (approved):
Yeah unfortunately the tests don't cover that... it's definitely an issue as it's not the first time that it happens unfortunately.

PR #134, opened 2023-08-04
Avoid serialisation breaking changes when possible

luckysori:
Obviously, there's not much we can do if an enum variant disappears or is renamed, or if the data associated with it changes. But we can usually avoid changing identifiers unnecessarily.

With this patch we ensure backwards compatibility in certain scenarios which were broken with 7560cad and 2013ad4.


Tibo-lg (approved):
Didn't think it would make a difference which is why I updated the numbers, but if it does fine by me.

luckysori:
> Didn't think it would make a difference which is why I updated the numbers, but if it does fine by me.

Awesome, thanks! We ran into this problem: https://github.com/get10101/10101/issues/1058. I also thought it wouldn't matter, but in this case we can avoid closing working LN channels with this patch.

luckysori:
@Tibo-lg: What steps do you follow to update the `sled` tests after a breaking change?

Tibo-lg:
> @Tibo-lg: What steps do you follow to update the `sled` tests after a breaking change?

```
$./scripts/generate_serialized_contract_files.sh 
```

It will run some integration tests and generate serialized files (requires docker + docker compose). If you feel like adding that to the docs it might be useful :)

PR #138, opened 2023-08-15
Feature/ln dlc/handle ldk auto closing

Tibo-lg:
This PR enables the `SubChannelManager` to react to force closing events from LDK. To achieve this, a couple of things were needed:
* Stop relying on LDK to sign transactions that are not LN specific (mainly makes things cleaner)
* Get ability to directly update the funding tx of the LDK channel monitor (through the LDK chain monitor)

Not all states are properly covered for now as it's not clear how to get LDK to trigger force closes in some of the states.


dlc-manager/src/sub_channel_manager.rs
```diff
@@ -1079,50 +1110,139 @@ where
...
+                    warn!("Got close notification while in ClosedPunished.");
+                    return Ok(());
+                }
+                SubChannelState::Rejected => {
+                    info!(
+                        "Counterparty closed channel in rejected state, marking as counter closed"
+                    );
+                    sub_channel.state = SubChannelState::CounterOnChainClosed;
+                    (None, None)
+                }
```
luckysori, line 1160: I reckon we could move this one to the branch with `OffChainClosed`. They're kind of equivalent: there is no subchannel.

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -1079,50 +1110,139 @@ where
...
+                            self.ln_chain_monitor
+                                .get_latest_holder_commitment_txn(&funding_txo)?,
+                        );
+                        self.force_close_signed_channel(sub_channel, commitment_transactions)?;
+                        return Ok(());
+                    }
+                    _ => return Ok(()),
+                }
             }
+            _ => return Ok(()),
```
luckysori, line 1245: I think it would be good to add the variants explicitly so that we don't miss handling important states if new variants are added as the code changes (including LDK!).

dlc-manager/src/subchannel/ser.rs
```diff
@@ -24,7 +24,8 @@ impl_dlc_writeable!(SubChannel, {
     (is_offer, writeable),
     (own_fund_pk, writeable),
     (counter_fund_pk, writeable),
-    (counter_party_secrets, writeable)
+    (counter_party_secrets, writeable),
+    (channel_keys_id, writeable)
```
luckysori, line 28: Note to self: this breaks all pre-existing channels.

dlc-manager/tests/ln_dlc_channel_execution_tests.rs
```diff
@@ -1604,6 +1616,190 @@ fn ln_dlc_close_finalized_force_close() {
...
+        get_commit_tx_from_node(&test_params.bob_node, &test_params.funding_txo).remove(0);
+
+    ldk_auto_close(&mut test_params, &commit_tx);
+}
+
+//TODO(tibo): find a way to test this state.
+// #[test]
+// #[ignore]
+// fn ln_dlc_ldk_auto_close_close_confirmed_test() {
+// }
```
luckysori, line 1767: For all the states for which force-close cannot be triggered with the fee estimate change, I would just call force-close on the LN channel directly?

luckysori (approved):
Nice work!

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -697,31 +719,47 @@ where
...
-        let mut sub_channel = self
+        let sub_channel = self
             .dlc_channel_manager
             .get_store()
             .get_sub_channel(*channel_id)?
             .ok_or(Error::InvalidParameters(format!(
                 "Unknown sub channel {:?}",
                 channel_id
             )))?;
+        let mut chain_monitor = self.dlc_channel_manager.get_chain_monitor().lock().unwrap();
```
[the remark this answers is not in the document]
Tibo-lg: Fair enough I had changed to this because of a deadlock issue but changed to grabbing it right before it's use in all places.

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -847,78 +889,63 @@ where
...
+                    sub_channel.fund_value_satoshis,
+                    sub_channel.channel_keys_id,
+                ),
+            );
+            let own_split_sig = signer.get_holder_split_tx_signature(
+                self.dlc_channel_manager.get_secp(),
+                &split_tx,
+                &sub_channel.original_funding_redeemscript,
+                sub_channel.fund_value_satoshis,
+            )?;
```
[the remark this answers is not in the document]
Tibo-lg: Yeah good point, added some helper methods let me know if that works for you

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -1079,50 +1110,139 @@ where
...
+                    chain_monitor.cleanup_channel(sub_channel.channel_id);
+                    sub_channel.state = SubChannelState::CounterOnChainClosed;
+                    (None, None)
+                }
+                SubChannelState::Signed(_)
+                | SubChannelState::Closing(_)
+                | SubChannelState::CloseOffered(_) => {
+                    error!("Got notification of LN channel closure by counter party in a state where we do not expect it.");
+                    return Ok(());
+                }
```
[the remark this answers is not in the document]
Tibo-lg: I need to spend some time remembering this. It indeed looks a bit strange.

Also changed to `Error`.

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -1079,50 +1110,139 @@ where
...
+                SubChannelState::OnChainClosed | SubChannelState::CounterOnChainClosed => {
+                    info!(
+                        "Channel close notification received for channel: {:?}",
```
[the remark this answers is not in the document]
Tibo-lg: Done

dlc-manager/src/sub_channel_manager.rs
Tibo-lg: Yeah but in the `OffChainClosed` state you need to cleanup the chain monitor, would be a bit weird to do it anyway for rejected channels I feel.

dlc-manager/src/sub_channel_manager.rs
Tibo-lg: I'm not convinced it's very useful but it doesn't hurt so added.

luckysori: It could be particularly useful in the case of a change in LDK (any dependency in general).

For instance, if we upgrade to 0.0.117 and they have added a new `ClosureReason` that we need to handle properly, the wildcard pattern would silently handle the new variant. Conversely, without the wildcard pattern the compiler would at least give us an error here and we would have to explicitly write code to handle the new variant.

dlc-manager/src/sub_channel_manager.rs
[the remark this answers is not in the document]
Tibo-lg: So I think the point is that these are states where the notification is probably not very useful:
* If you're in `Signed`, it means your counter party is as well, and if they close the split channel you'll get notified by finding the split tx on chain
* If you're in closing, it means you're already closing anyway so not much to do
* If you're in `CloseOffered` it's basically the same as `Signed`.
So I feel maybe just an `info!` message would be sufficient, but what do you think?

dlc-manager/tests/ln_dlc_channel_execution_tests.rs
Tibo-lg: I was hesitating to do that because it's somehow not something that you expect from proper usage. But well it's easier at the end so I just use that for all the tests now :)

dlc-manager/src/sub_channel_manager.rs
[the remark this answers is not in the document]
luckysori: Fair enough, I suppose it's enough to log for these and INFO level makes sense.

dlc-manager/src/sub_channel_manager.rs
luckysori: Ah, fair. Didn't notice the difference!

dlc-manager/src/sub_channel_manager.rs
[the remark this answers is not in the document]
Tibo-lg: Changed to log the channel id + state name as discussed.

Issue #132, opened 2023-07-27
Ensure that `CustomMessageHandler` impl is tested


luckysori:
_Originally posted by @Tibo-lg in https://github.com/p2pderivatives/rust-dlc/pull/131#pullrequestreview-1548998348_

PR #139, opened 2023-08-17
fix: Add missing renew revoke dlc message

holzeis:
Adds the missing `RenewRevoke` dlc message and fixes

```
lightning::ln::peer_handler[app]: Received unknown even message of type 43026, disconnecting peer!  
```

related to https://github.com/p2pderivatives/rust-dlc/issues/132


Tibo-lg (approved):
Thanks! (I guess we really need to resolve #132...)

PR #142, opened 2023-08-24
fix: Drop pending Finalize, Accept and CloseAccept messages on disconnected peer

holzeis:
I propose to drop the following messages after the counterparty disconnects.

- `Finalize`: After a reconnect before processing the finalize message, both parties will roll back to the `Offered` state. If the party is now trying to process the pending `Finalize` message an `InvalidState` error is reported. This is actually not critical and does not solve any bug, but rather cleans up the logs from a potentially misleading error message.
-  `Accept` & `CloseAccept`: If a reconnect happens in between receiving and processing the `Accept` or `CloseAccept` message, the signatures "break" on the accept message created in the context of the prior established channel, resulting into https://github.com/get10101/10101/issues/1141. This will even break the state in a way that subsequent `Accept` or `CloseAccept` message created in the context on the newly established channel will be successfully processed but fail subsequently on processing the `Finalize` or `FinalizeClose` message resulting into https://github.com/get10101/10101/issues/1074

I am uncertain if this is the right fix, or if the `Invalid state: Misuse error: Invalid commitment signed: Close : Invalid commitment tx signature from peer` shouldn't even happen in the first place. However, this fix resolves our issue as it prevents the scenario from happening.

### DIFF
--- a/dlc-messages/src/message_handler.rs
+++ b/dlc-messages/src/message_handler.rs
@@ -117,6 +117,16 @@ impl MessageHandler {
                     );
                     false
                 }
+                Message::SubChannel(SubChannelMessage::Finalize(message))
+                if node_id == disconnected_node_id =>
+                    {
+                        log::warn!(
+                        "Dropping SubChannelFinalize message for channel {:?} \
+                         after peer {node_id} disconnected",
+                        message.channel_id
+                    );
+                        false
+                    }
                 Message::SubChannel(SubChannelMessage::CloseConfirm(message))
                     if node_id == disconnected_node_id =>
                 {
@@ -127,6 +137,16 @@ impl MessageHandler {
                     );
                     false
                 }
+                Message::SubChannel(SubChannelMessage::CloseFinalize(message))
+                if node_id == disconnected_node_id =>
+                    {
+                        log::warn!(
+                        "Dropping SubChannelCloseFinalize message for channel {:?} \
+                         after peer {node_id} disconnected",
+                        message.channel_id
+                    );
+                        false
+                    }
                 // Keep any other message
                 _ => true,
             }
--- a/dlc-messages/src/message_handler.rs
+++ b/dlc-messages/src/message_handler.rs
@@ -137,16 +137,6 @@ impl MessageHandler {
                     );
                     false
                 }
-                Message::SubChannel(SubChannelMessage::CloseFinalize(message))
-                if node_id == disconnected_node_id =>
-                    {
-                        log::warn!(
-                        "Dropping SubChannelCloseFinalize message for channel {:?} \
-                         after peer {node_id} disconnected",
-                        message.channel_id
-                    );
-                        false
-                    }
                 // Keep any other message
                 _ => true,
             }
--- a/dlc-messages/src/message_handler.rs
+++ b/dlc-messages/src/message_handler.rs
@@ -137,6 +137,26 @@ impl MessageHandler {
                     );
                     false
                 }
+                Message::SubChannel(SubChannelMessage::Accept(message))
+                if node_id == disconnected_node_id =>
+                    {
+                        log::warn!(
+                        "Dropping SubChannelAccept message for channel {:?} \
+                         after peer {node_id} disconnected",
+                        message.channel_id
+                    );
+                        false
+                    }
+                Message::SubChannel(SubChannelMessage::CloseAccept(message))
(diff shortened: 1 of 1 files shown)


Tibo-lg:
Would it be difficult to add a test re-producing the issue before the fix?

luckysori:
As discussed offline, we are not merging this into `feature/ln-dlc-channels` for now.

holzeis:
@luckysori @Tibo-lg I managed to reproduce https://github.com/get10101/10101/issues/1141

```
Invalid state: Misuse error: Invalid commitment signed: Close : Invalid commitment tx signature from peer
```

I've added my changes to the original test in the following fixup https://github.com/get10101/10101/pull/1157/commits/2af11fb5b31d636c9f41a7ff0f8b33fc52ece799

It appears the error is happening through the following steps.

1. Reconnect when the `Finalize` message is received by the `coordinator`.
2. Both parties rollback to the `Offered` State upon a reestablished channel.
3. Process the pending `Finalize` message which will fail with `Invalid state: Expected Confirmed state but got Offered`
4. Call `process_actions()` on the apps `sub_channel_manager`, which will queue up the Accept message.
5. Reconnect when the `Accept` message is received by the `coordinator`.
6. Now once the channel is reestablished, the coordinator will fail on processing the accept message with the error `Invalid state: Misuse error: Invalid commitment signed: Close : Invalid commitment tx signature from peer`

Note, it looks like not processing the pending `Finalize` message in step 3 prevents the coordinator from running into the error in Step 6.

Tibo-lg:
@holzeis I tried to reproduce the steps you described sequentially here: https://github.com/p2pderivatives/rust-dlc/commit/4f334a7a0f1b5924bb4434098c592d9e25f7428b but the test doesn't fail (I can see the `Invalid state: Expected Confirmed state but got Offered` error but then it proceed sucessfully). I suspect due to your test having concurrency some race condition is happening. I think it might be processing the `Finalize` between the time the node sends its own `Reestablish` message and the time it receive the one from its peer. Could you check what is the state of both nodes before the error happens?

Tibo-lg:
Interestingly it seems to have uncover a different bug though 😅 
https://github.com/p2pderivatives/rust-dlc/actions/runs/5996063715/job/16260065184#logs

holzeis:
With https://github.com/get10101/10101/pull/1157/commits/3caa33c2f3e8527e712878628739f35e39ea7b5a I was also able to reproduce `Commitment txids are unique outside of fuzzing, where hashes can collide` which seems to be a follow up error after `Invalid state: Misuse error: Invalid commitment signed: Close : Invalid commitment tx signature from peer`

What needs to happen is that the app sens the correct accept message now, which pushes the protocol again to the finalize message which will then fail with the `Commitment txids are unique outside of fuzzing, where hashes can collide` error.

holzeis:
> @holzeis I tried to reproduce the steps you described sequentially here: [4f334a7](https://github.com/p2pderivatives/rust-dlc/commit/4f334a7a0f1b5924bb4434098c592d9e25f7428b) but the test doesn't fail (I can see the `Invalid state: Expected Confirmed state but got Offered` error but then it proceed sucessfully). I suspect due to your test having concurrency some race condition is happening. I think it might be processing the `Finalize` between the time the node sends its own `Reestablish` message and the time it receive the one from its peer. Could you check what is the state of both nodes before the error happens?

I had a quick look at your test and it looks like you are missing Step 5 in your test. A reconnect needs to happen in between receiving the `Accept` message and processing it. Something happens during the channel re-establishment that will break the `Accept` message producing `Invalid state: Misuse error: Invalid commitment signed: Close : Invalid commitment tx signature from peer`.

I am proposing a quick fix for that by simply dropping any `Accept` or `CloseAccept` message after the peer disconnects, but thats properly not the correct fix - as that should not happen either way?

holzeis:
@Tibo-lg: I cleaned up the tests a bit

This commit https://github.com/get10101/10101/commit/8241d97df1251240a54d785141415421b39e737c reproduces both errors 
 - Invalid state: Misuse error: Invalid commitment signed: Close : Invalid commitment tx signature from peer
 - Commitment txids are unique outside of fuzzing, where hashes can collide

This commit https://github.com/get10101/10101/commit/d5d04a7e055e6b8ee79543e167eeac741c07d18c points to the latest commit on this PR and fixes (prevents them from happening) those issues.

PR #150, opened 2023-09-18
Make Writeable trait public for oracle message

Pantamis:
Hi !

An oracle needs to sign the TLV encoding of the `OracleEvent` part of his announcement.

Because `Writeable` is declared as private, it seems to me impossible to access the `.encode()` or `write()` functions of an `OracleEvent` which is necessary to get the payload that will be signed with oracle's key to authenticate the announcement.

Maybe I am wrong and there is a simple way to use `.encode()` in an external crate, if not this means it is currently not possible to implement an oracle using `rust-dlc`.

This PR fixes it by declaring the use of the `Writeable` trait public.

### DIFF
--- a/dlc-messages/src/oracle_msgs.rs
+++ b/dlc-messages/src/oracle_msgs.rs
@@ -7,7 +7,8 @@ use crate::ser_impls::{
 use dlc::{Error, OracleInfo as DlcOracleInfo};
 use lightning::ln::msgs::DecodeError;
 use lightning::ln::wire::Type;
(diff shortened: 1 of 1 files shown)


Tibo-lg:
> Because Writeable is declared as private

Just import the `lightning` crate and use it from there. It's not declared in rust-dlc and I'm not sure re-exporting it is that useful.

Pantamis:
Update: I couldn't import `lightning` and use use the trait because of my `lightning` crate version, it must be set to 0.0.113 mine was 0.0.116. All good now :+1:

PR #141, opened 2023-08-22
Update to LDK v0.0.116

Tibo-lg:

PR #154, opened 2023-09-25
Use commit tx number to reject stale messages.

Tibo-lg:
This PR aims to fix the issue where stale messages are receive and mess up the internal state of the LDK node. It does so by embedding the commitment transaction number together with commitment signatures, and checking that we are given the expected one within the `with_useable_channel_lock` method.

I reverted previous changes to the message handler dropping messages on disconnect as I believe they are not necessary anymore with this change.

I can't promise that all concurrency issues will be resolved by this patch, but it should at least be more robust than it currently is.

(Leaving as draft for now as it builds on top of #141 which is not merged yet).

### DIFF
--- a/dlc-manager/src/sub_channel_manager.rs
+++ b/dlc-manager/src/sub_channel_manager.rs
@@ -483,6 +483,7 @@ where
         let msg = self.ln_channel_manager.with_useable_channel_lock(
             channel_id,
             &counter_party,
+            None,
             |channel_lock| {
                 let per_split_seed =
                     if let Some(per_split_seed_pk) = offered_sub_channel.per_split_seed {
@@ -615,7 +616,7 @@ where
                     .ln_chain_monitor
                     .get_latest_holder_commitment_txn(&funding_txo)?;
 
-                let commitment_signed = self
+                let (commitment_signed, commit_tx_number) = self
                     .ln_channel_manager
                     .get_updated_funding_outpoint_commitment_signed(
                         channel_lock,
@@ -668,6 +669,7 @@ where
                     publish_basepoint: offered_sub_channel.own_base_points.publish_basepoint,
                     own_basepoint: offered_sub_channel.own_base_points.own_basepoint,
                     commit_signature: commitment_signed.signature,
+                    commit_tx_number,
                     htlc_signatures: commitment_signed.htlc_signatures,
                     channel_revocation_basepoint: accept_channel.revocation_basepoint,
                     channel_publish_basepoint: accept_channel.publish_basepoint,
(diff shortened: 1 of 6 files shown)


dlc-manager/src/sub_channel_manager.rs
```diff
@@ -1772,12 +1779,13 @@ where
 
         let (
             split_tx_adaptor_signature,
-            commitment_signed,
+            (commitment_signed, commit_tx_number),
```
luckysori, line 1782: I guess this has to be the same number as the one that you pass in when calling `with_useable_channel_lock` just below.

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -2122,6 +2132,7 @@ where
         let msg = self.ln_channel_manager.with_useable_channel_lock(
             &sub_channel_finalize.channel_id,
             counter_party,
+            None,
```
luckysori, line 2135: How do we decide if we have to set this to `None`?

luckysori: I guess we set it to `Some` every time the received message includes a commitment signature which needs to be processed.

I think I should carefully review this file given this. It would be nicer if the check was included when processing a commitment signature to avoid programmer errors. But maybe the implementation would be a lot more complicated?

dlc-messages/src/message_handler.rs
```diff
@@ -102,38 +71,6 @@ impl MessageHandler {
         ret
     }
 
-    /// Drops certain pending messages sent by the remote node.
-    pub fn clear_pending_messages_sent_by_node(&self, disconnected_node_id: &PublicKey) {
```
luckysori, line 106: Nice!

dlc-manager/src/sub_channel_manager.rs
luckysori: I have reviewed this file carefully and it looks good!

luckysori (approved):
Nice! I think this is a step in the right direction.

dlc-manager/src/sub_channel_manager.rs
Tibo-lg: Mmm I guess it is the same but I don't know if there is a case where one party would maybe not yet have sent an RAA. That would probably mess things up though.

dlc-manager/src/sub_channel_manager.rs
Tibo-lg: I originally implemented it to check every time we pass a commitment signature, but the problem is that unfortunately the call to `get_updated_funding_outpoint_commitment_signed` is not properly reverted on error, causing the duplicate tx id error in the channel monitor. I think the reason is because `get_updated_funding_outpoint_commitment_signed` calls `monitor_updating_restored`. It might be possible to make things better by changing the API to return monitor updates and only committing the updates at the end (maybe even returning them from the `with_lock` method), but that'd be quite a big change. It definitely would be cleaner though.

dlc-manager/src/sub_channel_manager.rs
luckysori: Hmm, wouldn't it then fail when verifying the `commit_tx_number` provided by the caller?

Tibo-lg: Within the LN protocol it is possible for parties to have a different commitment number at a given time. So the one provided by the caller could potentially be different than the one received here (since the one provided is the one for the local commitment tx while the received one is for the remote commitment tx). But I doubt that we can reach a state where the number differ without first encountering other issues.

luckysori: Thanks for clarifying!

PR #156, opened 2023-10-17
feat: Add optional fee to be paid to the offering party

holzeis:
Adds an optional fee to be added to the payout function. This fee will be deducted from the accepting party payouts. This allows the offering party to charge for a fee his offering.

In 10101 we used to charge fees by automatically paying an invoice issued by the coordinator, which is not optimal as 

1. It is a trusted setup, where the coordinator trusts that the counter party pays the invoice.
2. It is pretty failure prune to immediately pay an invoice after updating the subchannel. We used to add a 5 seconds sleep to ensure everything is updated.

This change will allow us to charge fees (unilaterally) in an un-trusted way.

### DIFF
--- a/dlc-messages/src/contract_msgs.rs
+++ b/dlc-messages/src/contract_msgs.rs
@@ -178,9 +178,11 @@ pub struct PayoutFunction {
     pub payout_function_pieces: Vec<PayoutFunctionPiece>,
     /// The right most point of the function.
     pub last_endpoint: PayoutPoint,
+    /// An optional fee deducted by the offering party in the payouts.
+    pub fee: Option<u64>,
 }
 
-impl_dlc_writeable!(PayoutFunction, {(payout_function_pieces, vec), (last_endpoint, writeable)});
+impl_dlc_writeable!(PayoutFunction, {(payout_function_pieces, vec), (last_endpoint, writeable), (fee, option) });
 
 #[derive(Clone, Debug, PartialEq)]
 #[cfg_attr(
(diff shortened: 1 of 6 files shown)


Tibo-lg:
I'm ok with the concept of having a party pay a fee (we actually had that in the [cfd-dlc library](https://github.com/p2pderivatives/cfd-dlc/blob/develop/src/cfddlc_transactions.cpp#L113) originally but I didn't port it to rust-dlc). But I'm not convinced with your approach, it doesn't really make sense to me to have it as part of the payout function. I would put it on the `OfferedContract` probably because your approach means it'll only be available for numerical contracts, but someone might want the same functionality for enumerated contracts.

dlc-messages/src/contract_msgs.rs
```diff
@@ -178,9 +178,11 @@ pub struct PayoutFunction {
     pub payout_function_pieces: Vec<PayoutFunctionPiece>,
     /// The right most point of the function.
     pub last_endpoint: PayoutPoint,
+    /// An optional fee deducted by the offering party in the payouts.
+    pub fee: Option<u64>,
```
luckysori, line 182: Hmm, this is not what I had in mind. In fact I thought we would be able to do this without changing `rust-dlc` at all.

After experimenting a few weeks ago I concluded that you can apply the fee offset to the input (i.e. the margins) and it has the same effect as applying it on the output (i.e. the payouts).

holzeis:
Closing as @luckysori pointed out that this can be done without rust-dlc changes.

PR #158, opened 2023-10-19
chore: make function public

bonomat:
this is useful so that one can access create a spending tx from the funding tx


luckysori:
I verified that this is not needed since https://github.com/get10101/10101/blob/ee89af4b6018b7d875a06e4d9419b887602dcb29/coordinator/src/collaborative_revert.rs#L245-L259 does the trick.

PR #161, opened 2023-10-30
Implement ordinal support

Tibo-lg:
This PR adds the possibility to use an ordinal as part of the collateral of a contract. See `docs/Ordinal.md` for more details.


dlc-manager/src/contract/ord_descriptor.rs
```diff
@@ -0,0 +1,375 @@
...
+)]
+pub struct OrdDescriptor {
+    /// The descriptor for the contract outcomes.
+    pub outcome_descriptor: OrdOutcomeDescriptor,
+    /// The original location of the ordinal within the UTXO set.
+    pub ordinal_sat_point: SatPoint,
+    /// The transaction where the UTXO containing the ordinal is located.
+    pub ordinal_tx: Transaction,
+    /// Whether it the offer party that should be refunded in case of a contract timeout.
+    pub refund_offer: bool,
```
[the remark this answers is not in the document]
Tibo-lg: Probably could be better named, but this actually also indicates who provides the ordinal.

dlc-manager/src/contract/ord_descriptor.rs
```diff
@@ -0,0 +1,375 @@
...
+#[cfg_attr(
+    feature = "serde",
+    derive(Serialize, Deserialize),
+    serde(rename_all = "camelCase")
+)]
+pub struct OrdDescriptor {
+    /// The descriptor for the contract outcomes.
+    pub outcome_descriptor: OrdOutcomeDescriptor,
+    /// The original location of the ordinal within the UTXO set.
+    pub ordinal_sat_point: SatPoint,
```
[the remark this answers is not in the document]
Tibo-lg: `SatPoint` is the name used in the `ord` implementation: https://github.com/ordinals/ord/blob/master/src/sat_point.rs#L4

`output_index` seems like the index of the output within the transaction, but it's more than that since it also provides the offset for the actual satoshi.

dlc-manager/tests/ordinals_tests.rs
```diff
@@ -0,0 +1,702 @@
...
+
+fn get_enum_ord_outcome_descriptor(win_both: bool) -> OrdOutcomeDescriptor {
+    let outcome_payouts: Vec<_> = enum_outcomes()
+        .iter()
+        .enumerate()
+        .map(|(i, x)| {
+            let payout = if i % 2 == 0 {
+                Payout {
+                    offer: TOTAL_COLLATERAL,
+                    accept: 0,
```
sosaucily, line 140: This logic feels a bit fragile, here and down where creating the OrdEnumDescriptor to_offer_payouts. the i % 2 logic in both places takes some brainpower to follow and a bug could sneak it's way in.  maybe there's some way to hard-code the outcomes in a clearer way.

dlc-manager/src/contract/ord_descriptor.rs
```diff
@@ -0,0 +1,375 @@
...
+    derive(Serialize, Deserialize),
+    serde(rename_all = "camelCase")
+)]
+/// Descriptor for an event that has enumerated outcomes for contracts that involve an ordinal.
+pub struct OrdEnumDescriptor {
+    /// The description of the enumerated outcomes.
+    pub descriptor: EnumDescriptor,
+    /// The outcomes for which the ordinal is given to the offer party (one boolean for each
+    /// possible outcome).
+    pub to_offer_payouts: Vec<bool>,
```
sosaucily, line 93: Could we call this "to_offeror_payouts" ? to_offer makes it sound like it's mapping something to an Offer type or something, rather than the conditions when the ordinal goes to the offeror.

Tibo-lg: `offer` is what is used consistently in the code base and the DLC specifications so I'd rather not have it different just here.

dlc-manager/tests/ordinals_tests.rs
Tibo-lg: Yeah there are many ways to decide the payouts depending on the outcome. This one is just giving everything to one party for outcomes in odd position and the other party for outcomes in even positions. I thought it nice because it's not too simple while still not being too complicated IMHO (if we do e.g. the first half to one party and the other half to the other party it feels easier to miss some possible issues). But I don't have a strong opinion so let me know what you would prefer.

dlc-manager/src/contract/ord_descriptor.rs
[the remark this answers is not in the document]
Tibo-lg: Would something like `is_offer_owner` make more sense? Or do you have a better suggestion?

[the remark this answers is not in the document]
sosaucily: ah gotcha, i understand now. yes, that name would work better for me i think. and then the corresponding comment could be updated as well. If you'd like I can make this change and include it in a PR to this branch, or feel free to do it yourself.

dlc-manager/src/contract/ord_descriptor.rs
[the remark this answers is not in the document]
sosaucily: Ok, if that's how ORD does it, let's stick to the structure!

dlc-manager/tests/ordinals_tests.rs
sosaucily: OK, I'll review it more and think about whether I could like another design better. Maybe I'll propose something in a PR and we can discuss. Thanks for the feedback. Anyway, this isn't a blocker

dlc-manager/src/contract/ord_descriptor.rs
sosaucily: ok 👍

PR #162, opened 2023-10-31
fix: overflow bug in payout curve

bonomat:
It is possible that a payout curve decreases with the amount of payout as the events increase. This lead to an overflow panic in the evaluate function

### DIFF
--- a/dlc-manager/src/payout_curve.rs
+++ b/dlc-manager/src/payout_curve.rs
@@ -311,7 +311,7 @@ impl Evaluable for PolynomialPayoutCurvePiece {
             return if left_point.outcome_payout == right_point.outcome_payout {
                 right_point.outcome_payout as f64
             } else {
-                let slope = (right_point.outcome_payout - left_point.outcome_payout) as f64
+                let slope = (right_point.outcome_payout as f64 - left_point.outcome_payout as f64)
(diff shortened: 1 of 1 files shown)


dlc-manager/src/payout_curve.rs
```diff
@@ -311,8 +311,8 @@ impl Evaluable for PolynomialPayoutCurvePiece {
             return if left_point.outcome_payout == right_point.outcome_payout {
                 right_point.outcome_payout as f64
             } else {
-                let slope = (right_point.outcome_payout - left_point.outcome_payout) as f64
-                    / (right_point.event_outcome - left_point.event_outcome) as f64;
+                let slope = (right_point.outcome_payout as f64 - left_point.outcome_payout as f64)
+                    / (right_point.event_outcome as f64 - left_point.event_outcome as f64);
```
Tibo-lg, line 315: nit: as we expect event outcomes to be increasing only you don't really need to change this line

Tibo-lg (approved):
Good catch, surprised that it was not caught before.

holzeis (approved):

dlc-manager/src/payout_curve.rs
bonomat: Thanks for pointing this out. I've removed these changes.

PR #164, opened 2023-11-01
Update rust-lightning dependency

luckysori:
The backwards-compatibility problem is associated with the recent upgrade to `rust-lightning:0.0.116`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,6 @@ members = [
 resolver = "2"
 
 [patch.crates-io]
-lightning = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "b13daab9" }
-lightning-net-tokio = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "b13daab9" }
-lightning-persister = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "b13daab9" }
+lightning = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "a57281b" }
+lightning-net-tokio = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "a57281b" }
+lightning-persister = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "a57281b" }


luckysori:
@Tibo-lg: This PR syncs up our branch with `split-tx-experiment` in LDK now.

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -316,6 +316,11 @@ where
                     SubChannelState::OffChainClosed => {
                         s.is_offer = true;
                         s.update_idx -= 1;
+
+                        // We do this to give older channels a chance to upgrade to the new version which
```
Tibo-lg, line 320: Honestly I'm not a big fan of this. It feels hacky and error prone. I'd like to have a way to handle upgrading at de-serialization time. Not sure what the best way is to be honest, but for example LDK uses the `ReadableArgs` trait for when deserialization requires external data.

luckysori: Yeah, it was supposed to be hacky in this case.

We don't have to merge this here anyway. I will just update the PR and we can just keep depending on this patch in 10101.

Tibo-lg: If you think it's time to take backward compatibility issues seriously, it's probably worth taking the time to find a proper solution (though it's also hard to say how many more breaking changes will be required to reach a stable implementation).

luckysori: I agree with you on needing to figure out how to do this properly as soon as possible. But _we_ can probably live with this quick fix for now.

dlc-messages/src/ser_impls.rs
```diff
@@ -453,7 +453,14 @@ pub fn read_option_cb<R: ::std::io::Read, T, F>(
 where
     F: Fn(&mut R) -> Result<T, DecodeError>,
 {
-    let prefix: u8 = Readable::read(reader)?;
+    let prefix: u8 = match Readable::read(reader) {
```
Tibo-lg, line 456: I wonder if it'd be better to have a different option type that allows the field not being present?

Tibo-lg (commented):
I think you can drop the first commit.

dlc-messages/src/ser_impls.rs
luckysori: Maybe! I thought this patch wouldn't be controversial, but it's fair to say that we don't need it since we dropped the other one. I'll remove this one too!

Tibo-lg (approved):

PR #165, opened 2023-11-01
Update to tip of `p2pderivatives/rust-lightning#split-channel-experiment-116`

luckysori:
Nothing new other than updating to the tip of `p2pderivatives/rust-lightning#split-channel-experiment-116`, instead
of depending on a merged PR branch.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,6 @@ members = [
 resolver = "2"
 
 [patch.crates-io]
-lightning = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "b13daab9" }
-lightning-net-tokio = { git = "https://github.com/p2pderivatives/rust-lightning/", rev = "b13daab9" }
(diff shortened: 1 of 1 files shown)


Tibo-lg:
Shouldn't we be using the tip of `p2pderivatives/rust-lightning#split-channel-experiment`? I think they're roughly the same but if we're to use it as out "main" branch it might be better?

luckysori:
> Shouldn't we be using the tip of `p2pderivatives/rust-lightning#split-channel-experiment`? I think they're roughly the same but if we're to use it as out "main" branch it might be better?

Yep, we want to use the tip of `rust-lightning#split-tx-experiment`. This PR was an intermediary one which updated to an _older_ tip of said branch. But now we can just close it and look at https://github.com/p2pderivatives/rust-dlc/pull/164 instead. That one is only one patch behind the tip, but I can update that quickly as the patch is inconsequential for consumers.

PR #166, opened 2023-11-02
Load ChainMonitor in (DLC) Manager constructor if possible

luckysori:
Otherwise we forget about all the registered transactions on restart, which can lead to loss of funds.


Tibo-lg:
That's quite a bad oversight thanks! Would love to have a test for it but I can take care of it later.

PR #168, opened 2023-11-03
Load `ChainMonitor` in (DLC) `Manager` constructor if possible

luckysori:
Otherwise we forget about all the registered transactions on restart, which can lead to loss of funds.

---

With this PR we are "backporting" a [similar patch](https://github.com/p2pderivatives/rust-dlc/pull/166) from the `feature/ln-dlc-channels` branch.


Tibo-lg (approved):
Thanks!

PR #175, opened 2023-11-20
Update documentation for ordinals

sosaucily:


docs/Ordinal.md
```diff
@@ -1,13 +1,41 @@
...
 
-It is possible to use an ordinal as part of the collateral of a DLC.
-To do so, an [`OrdDescriptor`](../dlc-manager/src/contract/ord_descriptor.rs) must be used.
-This descriptor contains information about the ordinal being included in the DLC (location in the blockchain and the transaction that includes is), as well as information about the event upon which the DLC is based.
+This library now supports the locking of an ordinal as part of the collateral of a DLC.
+
+Notes:
+* An ordinal is inscribed into a single satoshi. However, that satoshi is often part of a larger UTXO. DLCs lock UTXOs, so please keep that context in mind.
+
+* In all cases, the gas fee to lock and later unlock the DLC is split between both parties.
```
yancyribbens, line 8: gas fee is eth terminology.  I think you mean transaction fee.

docs/Ordinal.md
```diff
@@ -1,13 +1,41 @@
 # Ordinal Support
 
-It is possible to use an ordinal as part of the collateral of a DLC.
-To do so, an [`OrdDescriptor`](../dlc-manager/src/contract/ord_descriptor.rs) must be used.
-This descriptor contains information about the ordinal being included in the DLC (location in the blockchain and the transaction that includes is), as well as information about the event upon which the DLC is based.
+This library now supports the locking of an ordinal as part of the collateral of a DLC.
```
Tibo-lg, line 3:
```suggestion
This library supports the locking of an ordinal as part of the collateral of a DLC.
```

docs/Ordinal.md
```diff
@@ -1,13 +1,41 @@
...
+
+Notes:
+* An ordinal is inscribed into a single satoshi. However, that satoshi is often part of a larger UTXO. DLCs lock UTXOs, so please keep that context in mind.
```
Tibo-lg, line 6: I'm not an expert, but I don't think an ordinal is incribed. Some content can be "inscribed" into a single stats, but AFAIU an ordinal is the number associated with the sats.

Tibo-lg: Also I'm not sure what you're trying to say here tbh.

docs/Ordinal.md
```diff
@@ -1,13 +1,41 @@
...
+
+* In all cases, the gas fee to lock and later unlock the DLC is split between both parties.
+
+* For more information about DLCs, please see the main folder of this repo, the [`DLC-Link docs page`](https://docs.dlc.link), https://github.com/dlc-link, and the main https://github.com/discreetlogcontracts/dlcspecs specifications page.
```
Tibo-lg, line 10: I don't think that's relevant here, this is not supposed to be a standalone document.

docs/Ordinal.md
```diff
@@ -1,13 +1,41 @@
...
+
+> Resizing UTXOs
+>
+> Please note, this package does not currently support the re-sizing of the UTXO containing the ordinal within the DLC. See [this section](#changing-postage) for more details
```
Tibo-lg, line 14: Why use `>` here?

Tibo-lg: Also since there is already a section about that under I feel this is redundant.

docs/Ordinal.md
```diff
@@ -1,13 +1,41 @@
...
+### DLC with only the ordinal
+This package supports the use case when the offering party of the DLC locks the ordinal-containing UTXO into the DLC, and the counter-party (the acceptor) does not lock any collateral.
+
+In this case, the outcome is essentially binary, either the ordinal containing UTXO goes to the offeror, or the acceptor.
+
+### Ordinal with additional collateral
+This package also supports the case when the offering party of the DLC locks the ordinal-contianing UTXO into the dlc, AND any amount of addition collateral is locked into the DLC by either party.
+
+1. In this case, the outcome of the UTXO containing the ordinal will go to either party A or B, without resizing (as described earlier)
+2. The remaining collateral locked by either, or both, parties can be split between the parties based on the standard DLC enumerated or numerical outcome guidelines.
```
Tibo-lg, line 28: I feel this could be summarized in a sentence saying that you can chose to attach an additional collateral or not. To me it's quite difficult to grasp as is.

docs/Ordinal.md
```diff
@@ -1,13 +1,41 @@
...
+
+### Locking an ordinal in the DLC
```
Tibo-lg, line 30: Should that be part of the `Supported types of DLC` section?

Tibo-lg (changes requested):

docs/Ordinal.md
sosaucily: Well I was writing this from the perspective of someone who isn't fully versed in DLCs. So for that I was being more verbose and describing exactly what is happening in the DLC. Since there is no reference to Ordinals in the DLC specs anywhere, this can be used as the informational page. DLC Specs covers a lot of this high-level text. 

If you'd like however, I can shrink it back down, and find another place, perhaps on our dlc.link docs, to add this kind of less-technical review of ordinals and DLCs.

docs/Ordinal.md
sosaucily: Ah, probably not. Let me move stuff around a bit.

docs/Ordinal.md
sosaucily: OK, removed

docs/Ordinal.md
sosaucily: Again I'm trying to give context for the whole effort. However, if you prefer to keep the document closer to as you had it before, I'll move some of this conversation outside to our own docs page. What do you think?

docs/Ordinal.md
sosaucily: removed this whole section

docs/Ordinal.md
Tibo-lg: Yeah maybe that's better, I don't think this is the place where people should learn about ordinals tbh (I'd expect that someone who wants to use this feature of the library would already know about it).

docs/Ordinal.md
Tibo-lg: I think it's important to have information about how the library can be used, but somehow I'm not seeing these sentences as very helpful. I think what would be most helpful would be to have some [examples](https://doc.rust-lang.org/cargo/reference/cargo-targets.html#examples) if you feel motivated.

docs/Ordinal.md
```diff
@@ -1,33 +1,46 @@
...
+
+## Ordinals & additional collateral
+
+The DLC can contain only 1 ordinal, either locked by the offering party or the accepting party.
+
+Either of the participants can choose to include additional satoshis as collateral into the DLC, along with the ordinal.
+
+## How it works
+
+In order to ensure that the ordinal does not get lost to fee, the DLC transactions for DLCs including an ordinal are created in the following way:
```
Tibo-lg, line 13:
```suggestion
In order to ensure that the ordinal gets properly assigned and not given to miners as fee, the DLC transactions for DLCs including an ordinal are created in the following way:
```

docs/Ordinal.md
```diff
@@ -1,33 +1,46 @@
...
+
+- The ordinal input is always set as the first one in the funding transaction,
+- The funding output is always set as the first one in the funding transaction,
+- The party getting the ordinal will always have its CET output in the first position.
+
+### Locking an ordinal in the DLC
```
Tibo-lg, line 19:
```suggestion
### Locking an ordinal in a DLC
```

Tibo-lg (commented):
I'm mostly fine with it now, just a couple of suggestions.

docs/Ordinal.md
sosaucily: I agree. If someone's so deep in the rust-dlc code, they should have some background on the various technologies which brought them there.

docs/Ordinal.md
sosaucily: That's cool. As I try out the library on our dlc.link system i'll see if I can add some examples

sosaucily:
Can commit suggestions directly from the github UI!

Tibo-lg (approved):
Thanks!

yancyribbens:
It looks like the term gas fee is still being used.  Is this intentional?

Tibo-lg:
> It looks like the term gas fee is still being used. Is this intentional?

I don't see it personally

yancyribbens:
oops, I must have been looking at an outdated diff.

sosaucily:
> oops, I must have been looking at an outdated diff.

Thanks for taking the time to give a review @yancyribbens.

PR #176, opened 2023-11-22
Update to `rust-lightning:0.0.117`

luckysori:
The highlight in terms of code changes in `rust-dlc` is:

Use new LDK Channel ID struct
=============================

Start to use newly introduced `ChannelId` struct. With a stronger type we are less likely to mix up LN `ChannelId`s, DLC channel IDs and contract IDs. To that end, we've renamed the old `ChannelId` type alias to `DlcChannelId` since it is now only supposed to be used for (temporary or final) DLC channel IDs. We should consider introducing a struct at some point.

To ensure backwards-compatibility we're still serialising the LN channel IDs into bytes, but we now have to use the `serialize_with` attribute since LDK's `ChannelId` doesn't implement serde itself. The same applies to deserialisation.


Tibo-lg:
> #### Use new LDK Channel ID struct
> Start to use newly introduced `ChannelId` struct. With a stronger type we are less likely to mix up LN `ChannelId`s, DLC channel IDs and contract IDs. To that end, we've renamed the old `ChannelId` type alias to `DlcChannelId` since it is now only supposed to be used for (temporary or final) DLC channel IDs. We should consider introducing a struct at some point.
> 
> To ensure backwards-compatibility we're still serialising the LN channel IDs into bytes, but we now have to use the `serialize_with` attribute since LDK's `ChannelId` doesn't implement serde itself. The same applies to deserialisation.

I feel it'd be cleaner to just define our own structs, and implement `From<ldk::ChannelId>` for the `SubChannelId` one. Then we could derive the `serde::Serialize`, and also have a more consistent naming. I feel it can be confusing that `Channel` structs use `DlcChannelId` while `SubChannel` use `ChannelId`. Unless there is some advantage that I'm not seeing?

luckysori:
> > #### Use new LDK Channel ID struct
> > Start to use newly introduced `ChannelId` struct. With a stronger type we are less likely to mix up LN `ChannelId`s, DLC channel IDs and contract IDs. To that end, we've renamed the old `ChannelId` type alias to `DlcChannelId` since it is now only supposed to be used for (temporary or final) DLC channel IDs. We should consider introducing a struct at some point.
> > To ensure backwards-compatibility we're still serialising the LN channel IDs into bytes, but we now have to use the `serialize_with` attribute since LDK's `ChannelId` doesn't implement serde itself. The same applies to deserialisation.
> 
> I feel it'd be cleaner to just define our own structs, and implement `From<ldk::ChannelId>` for the `SubChannelId` one. Then we could derive the `serde::Serialize`, and also have a more consistent naming. I feel it can be confusing that `Channel` structs use `DlcChannelId` while `SubChannel` use `ChannelId`. Unless there is some advantage that I'm not seeing?

I think what you propose is even better, but I wasn't sure if you wanted to go that far since before this PR we were just using type aliases everywhere. I can work on adding a couple of newtypes like you suggest.

luckysori:
@Tibo-lg: I've tried to address your [comment](https://github.com/p2pderivatives/rust-dlc/pull/176#issuecomment-1825043623).

It's probably not perfect because it was a pretty painful process, so let me know if I can make it better.

I had to move some stuff around so that the new structs could be used in all the necessary places. I also left a couple of TODOs and a NOTE. Let me know what you think about those too.

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -228,10 +231,11 @@ where
             ln_channel_manager,
             dlc_channel_manager,
             actions: Mutex::new(actions),
-            phantom: PhantomData,
+            phantom_cs: PhantomData,
```
Tibo-lg, line 234: I think you don't need the `CS` generic parameter anymore.

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -359,8 +363,7 @@ where
         let per_split_seed_pk =
             PublicKey::from_secret_key(self.dlc_channel_manager.get_secp(), &per_split_seed);
 
-        let temporary_channel_id: ContractId =
-            subchannel::generate_temporary_channel_id(*channel_id, update_idx, 0);
+        let temporary_channel_id = generate_temporary_dlc_channel_id(channel_id, update_idx, 0);
```
Tibo-lg, line 366: Feels like this would be better as a method on the type.

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -2026,7 +2039,8 @@ where
                 )?;
 
                 let sign_channel = dlc_messages::channel::SignChannel {
-                    channel_id: sub_channel_confirm.channel_id,
+                    // NOTE: I think this was wrong before!
```
Tibo-lg, line 2042: I think you're right. It didn't show up because this field is not used in `verify_signed_contract_internal` where this is passed.

dlc-manager/src/sub_channel_manager.rs
```diff
@@ -2899,9 +2917,11 @@ where
             let mut sub_channel = match self
                 .dlc_channel_manager
                 .get_store()
-                .get_sub_channel(channel_info.channel_id)?
+                .get_sub_channel(SubChannelId::from_bytes(channel_info.channel_id))?
             {
                 None => {
+                    // TODO: An error might be too much, especially since we will get
+                    // `DlcChannelId`s here even though they are supposed to be handled elsewhere!
```
Tibo-lg, line 2924: I don't get what you're saying here.

dlc-manager/src/subchannel/mod.rs
```diff
@@ -395,9 +408,9 @@ pub struct ClosingSubChannel {
 }
 
 /// Provides the ability to access and update Lightning Network channels.
-pub trait LNChannelManager<SP>: ChannelMessageHandler
+pub trait LNChannelManager<K: Deref>: ChannelMessageHandler
```
Tibo-lg, line 411: nit: feels like `SP` was a pretty good generic id for `SignerProvider`?

dlc-manager/tests/ln_dlc_channel_execution_tests.rs
```diff
@@ -259,26 +261,28 @@ impl TestScorer {
...
     fn channel_penalty_msat(
         &self,
-        _short_channel_id: u64,
-        _source: &NodeId,
-        _target: &NodeId,
-        _usage: ChannelUsage,
+        _: u64,
+        _: &NodeId,
+        _: &NodeId,
+        _: ChannelUsage,
```
Tibo-lg, line 271: not sure why you changed these and not the last one :D (I feel not changing them is also fine)

dlc-manager/tests/test_utils.rs
```diff
@@ -714,3 +715,59 @@ pub fn refresh_wallet<B: Deref, W: Deref>(
         retry += 1;
     }
 }
+
+/// Serialize an hexadecimal value.
+pub fn serialize_hex<S>(hex: &[u8], s: S) -> Result<S::Ok, S::Error>
```
Tibo-lg, line 720: I think there are the same methods in `dlc::serde_utils`, maybe easier to just reference that?

dlc-messages/Cargo.toml
```diff
@@ -12,8 +12,8 @@ use-serde = ["serde", "secp256k1-zkp/use-serde"]
 
 [dependencies]
 bitcoin = {version = "0.29.2"}
-dlc = {version = "0.4.0", path = "../dlc"}
-lightning = {version = "0.0.116"}
+dlc = {version = "0.4.0", path = "../dlc", features = ["serde"] }
```
Tibo-lg, line 15: I think `serde` is supposed to be feature gated (not 100% sure it's correctly setup but at least that was the goal)

dlc/Cargo.toml
```diff
@@ -9,10 +9,11 @@ version = "0.4.0"
 
 [dependencies]
 bitcoin = {version = "0.29.2"}
+lightning = {version = "0.0.117"}
```
Tibo-lg, line 12: I would rather this crate to stay very simple and not import `lightning` unless necessary. Actually is there a reason that you put `ids.rs` here and not in `dlc-manager`? It doesn't seem to be used within this crate.

mocks/src/memory_storage_provider.rs
```diff
@@ -11,19 +12,19 @@ use dlc_manager::contract::{
...
 use secp256k1_zkp::{PublicKey, SecretKey};
 use simple_wallet::WalletStorage;
 use std::collections::HashMap;
 use std::sync::{Mutex, RwLock};
 
 pub struct MemoryStorage {
     contracts: RwLock<HashMap<ContractId, Contract>>,
-    channels: RwLock<HashMap<ChannelId, Channel>>,
-    sub_channels: RwLock<HashMap<ChannelId, SubChannel>>,
+    channels: RwLock<HashMap<[u8; 32], Channel>>,
```
Tibo-lg, line 23: Is there a reason for not using the types here?

Tibo-lg (commented):
Overall LGTM, still a couple of things that need fixing.

dlc-manager/src/manager.rs
```diff
@@ -130,8 +131,8 @@ macro_rules! check_for_timed_out_channels {
                 if is_timed_out {
                     let sub_channel = if channel.is_sub_channel() {
                         log::info!(
-                            "Skipping force-closure of subchannel {}: not supported",
-                            bitcoin::hashes::hex::ToHex::to_hex(&channel.channel_id[..])
+                            "Skipping force-closure of subchannel {:?}: not supported",
```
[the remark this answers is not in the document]
luckysori: I agree, we need to print this as hex. I'll fix it.

dlc-manager/src/manager.rs
```diff
@@ -787,6 +788,9 @@ where
             .map(|x| self.get_oracle_announcements(&x.oracles))
             .collect::<Result<Vec<_>, Error>>()?;
 
+        let channel_id = crate::utils::get_new_temporary_id();
+        let channel_id = DlcChannelId::from_bytes(channel_id);
```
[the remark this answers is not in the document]
luckysori: Yeah, I agree. I will do that.

dlc-manager/src/manager.rs
```diff
@@ -1756,7 +1763,7 @@ where
         self.chain_monitor.lock().unwrap().add_tx(
             prev_tx_id,
             ChannelInfo {
-                channel_id: signed_channel.channel_id,
+                channel_id: signed_channel.channel_id.inner(),
```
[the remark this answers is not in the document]
luckysori: Yep. I think `From`/`Into` makes sense. Can do that.

dlc-manager/src/sub_channel_manager.rs
luckysori: My understanding is that `confirmed_txs` will return confirmed transactions for both subchannels and DLC channels because they're registered based on:

```rust
pub(crate) struct ChannelInfo {
    /// The identifier for _either_ a Lightning channel or a DLC channel.
    pub channel_id: [u8; 32],
    pub tx_type: TxType,
}
```

So if a transaction registered for a DLC channel (based on a `DlcChannelId`) is confirmed and has not been cleaned up yet (via `cleanup_channel`), it will pop up here and we will not be able to find a subchannel in the store for it, which will lead to us printing this error unnecessarily.

Alternatively we could introduce a `ChannelKind` enum in `chain_monitor.rs` so that the caller can, for example, filter out the confirmed transactions for irrelevant channel kinds. That way we can avoid misleading log messages.

dlc-manager/src/subchannel/mod.rs
luckysori: Good catch! I think this is a remnant of a previous version that involved more type parameters.

dlc-manager/tests/ln_dlc_channel_execution_tests.rs
luckysori: Yeah, that's weird. Definitely not intentional. I'll revert it.

dlc-manager/tests/test_utils.rs
luckysori: Yep, will do that!

dlc-messages/Cargo.toml
luckysori: You are right. We want to only enable `serde` for the `dlc` crate if the `serde` feature flag is passed. I'll fix that.

dlc/Cargo.toml
luckysori: > Actually is there a reason that you put ids.rs here and not in dlc-manager? It doesn't seem to be used within this crate.

I was also hesitant about this. But it seemed to me like IDs belong in a core crate.

The reason why I didn't want to put them in `dlc-manager` was because IDs are used in `dlc-messages`, meaning that `dlc-messages` would have to depend on `dlc-manager` which seems weird. And, more importantly, that would create a circular dependency.

> I would rather this crate to stay very simple and not import lightning unless necessary.

I considered adding a feature flag to the `dlc` crate, but it was getting a bit tedious so I dropped it. Should I try that again?

mocks/src/memory_storage_provider.rs
luckysori: Hehe, I had the types here until the very last second, when the compiler started giving me a hard time about the types not implementing certain traits. I can give it a go again.

dlc/Cargo.toml
Tibo-lg: Yeah a circular dependency wouldn't be good. What do you think about just not using these types for the `dlc-messages` crate? I guess it's not that a big deal to have it in `dlc` though, so I'll leave it up to you.

dlc-manager/src/sub_channel_manager.rs
Tibo-lg: Ok got it thanks! I feel a proper fix would be to have each manager having its own channel monitor, but being updated using the same data (so for example you'd only process blocks once).

luckysori: I totally agree with you comment, but I guess such a fix doesn't belong in this PR.

I will just remove the TODO and create an issue that we can address later.

luckysori: Issue here: https://github.com/p2pderivatives/rust-dlc/issues/193.

PR #182, opened 2023-12-04
return counterparty pk when rejecting subchannel offer and handle rejected state

bonomat:
Useful because we will need to send the reject message afterwards and for that we need to know to whom.

### DIFF
--- a/dlc-manager/tests/ln_dlc_channel_execution_tests.rs
+++ b/dlc-manager/tests/ln_dlc_channel_execution_tests.rs
@@ -2860,7 +2860,7 @@ fn reject_offer(test_params: &LnDlcTestParams) {
         Offered
     );
 
-    let reject = test_params
+    let (_, reject) = test_params
         .bob_node
         .sub_channel_manager
         .reject_sub_channel_offer(test_params.channel_id)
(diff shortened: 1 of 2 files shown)


Tibo-lg (changes requested):
LGTM but I'd like at least one integration test to make sure the re-opening of a rejected channel is working as expected.

bonomat:
> LGTM but I'd like at least one integration test to make sure the re-opening of a rejected channel is working as expected.

I've added a test https://github.com/p2pderivatives/rust-dlc/compare/81ac40c551203a18be147feca7e4361d02ef2567..e3c3602bdd6d32fd00dbff2577671269ab61cb50

dlc-manager/tests/ln_dlc_channel_execution_tests.rs
```diff
@@ -951,6 +951,29 @@ fn ln_dlc_rejected_offer() {
...
+
+#[test]
+#[ignore]
+fn ln_dlc_rejected_offer_then_new_offer_and_accept_offer() {
+    let test_params = test_init();
+    make_ln_payment(&test_params.alice_node, &test_params.bob_node, 900000);
+
+    reject_offer(&test_params);
+
+    offer_sub_channel_internal(&test_params, false);
```
Tibo-lg, line 963: nit: can you use `open_sub_channel` instead since it's there?

bonomat: done: https://github.com/p2pderivatives/rust-dlc/pull/182/commits/5f80d389a038baa3d4210625cfb1a63f3a3e7c4b

bonomat:
I'll squash the last two commits before merge

Tibo-lg (approved):
Thanks! LGTM

PR #189, opened 2023-12-12
Feat/add protocol fee

sosaucily:


dlc/src/lib.rs
```diff
@@ -400,20 +449,65 @@ pub(crate) fn create_fund_transaction_with_fees(
...
+            ),
+            _ => (
+                0,
+                Address::from_str("1111111111111111111114oLvT2").expect("valid address"),
+            ),
+        };
+
+    println!("desired value for fee: {:?}", (protocol_fee_amount));
+    let acceptor_protocol_fee_output = TxOut {
+        value: protocol_fee_amount,
```
bonomat, line 484: I think you need to check if the fee is above dust.

dlc-manager/src/contract/offered_contract.rs
```diff
@@ -44,6 +44,11 @@ pub struct OfferedContract {
     pub cet_locktime: u32,
     /// The time at which the contract becomes refundable.
     pub refund_locktime: u32,
+    /// The denominator `x` for i = `1/x`, where i is the percentage of locked collateral to be paid to the protocol as a fee.
+    /// 0 for no fee
+    pub fee_percentage_denominator: u64,
+    /// The address to which the protocol fee is paid
+    pub fee_address: String,
```
bonomat, line 51: Might be better to use `Address` here to fail early if the provided address is not valid

sosaucily: Thanks @bonomat, I will try, but I think I needed to use a String at this level because of the way the `impl_dlc_writeable` macro works, and it needs lightning::util::ser::Writeable, which Address does not.

dlc/src/lib.rs
sosaucily: I think the outputs under the dust limit get filtered later in the function, under the create_funding_transaction function.

sosaucily:
Thanks @bonomat for the comments! I accidentally opened this PR against the p2pderivatives repo, rather than my own fork. So I'll close this, but I appreciate you jumping in with some feedback! Let me know if you have any questions or comments!

bonomat:
> Thanks @bonomat for the comments! I accidentally opened this PR against the p2pderivatives repo, rather than my own fork. So I'll close this, but I appreciate you jumping in with some feedback! Let me know if you have any questions or comments!

No worries, feel free to ping me if you want further feedback on this.

PR #194, opened 2024-01-17
Fix/payout curve bugs

Tibo-lg:
This PR ports a couple of fixes that were applied to the `ln-dlc-channels` branch but not master.


luckysori (approved):
Thanks for porting these!

PR #196, opened 2024-01-30
fix: Add missing reject message

holzeis:
Adds the missing `Reject` dlc message and fixes

```
lightning::ln::peer_handler[app]: Received unknown even message of type 43024, disconnecting peer!  
```

related to https://github.com/p2pderivatives/rust-dlc/issues/132


Tibo-lg (approved):
Thanks! I guess we should indeed address #132 at some point.

PR #198, opened 2024-01-31
fix: Set timeout on `RenewOffered` state

holzeis:
Otherwise the channel gets force closed immediately if the periodic check happens to run before the channel has advanced to the next state.


Tibo-lg (approved):
Thanks! Would be nice to have a test but I'll try to add one later.

PR #199, opened 2024-02-01
feat: Add reject dlc channel offer handling

holzeis:
This change allows the counterparty to reject a dlc channel offer and adds a handling to gracefully transition the dlc channel to a `Cancelled` state.

Not super critical, but IMHO it makes it a bit clearer if the dlc channel offer is stuck or rejected.

### DIFF
--- a/dlc-manager/src/manager.rs
+++ b/dlc-manager/src/manager.rs
@@ -20,6 +20,7 @@ use crate::error::Error;
 use crate::{ChannelId, ContractId, ContractSignerProvider};
 use bitcoin::locktime::Height;
 use bitcoin::Transaction;
+use bitcoin::hashes::hex::ToHex;
 use bitcoin::{locktime, Address, LockTime};
 use dlc_messages::channel::{
     AcceptChannel, CollaborativeCloseOffer, OfferChannel, Reject, RenewAccept, RenewConfirm,
@@ -922,6 +923,17 @@ where
         )?;
 
         Ok(msg)
+    }
+
+    /// Reject a channel that was offered. Returns the [`dlc_messages::channel::Reject`]
+    /// message to be sent as well as the public key of the offering node.
+    pub fn reject_channel(&self, channel_id: &ChannelId) -> Result<(Reject, PublicKey), Error> {
+        let offered_channel = get_channel_in_state!(self, channel_id, Offered, None as Option<PublicKey>)?;
+        let counterparty = offered_channel.counter_party;
+        self.store.upsert_channel(Channel::Cancelled(offered_channel), None)?;
+
+        let msg = Reject{ channel_id: *channel_id };
+        Ok((msg, counterparty))
     }
 
     /// Accept a channel that was offered. Returns the [`dlc_messages::channel::AcceptChannel`]
@@ -1986,14 +1998,39 @@ where
         Ok(())
     }
 
-    fn on_reject(&mut self, reject: &Reject, counter_party: &PublicKey) -> Result<(), Error> {
-        let mut signed_channel =
-            get_channel_in_state!(self, &reject.channel_id, Signed, Some(*counter_party))?;
+    fn on_reject(&self, reject: &Reject, counter_party: &PublicKey) -> Result<(), Error> {
+        let channel = self.store.get_channel(&reject.channel_id)?;
 
-        crate::channel_updater::on_reject(&mut signed_channel)?;
+        if let Some(channel) = channel {
+            if channel.get_counter_party_id() != *counter_party {
+                return Err(Error::InvalidParameters(format!(
+                    "Peer {:02x?} is not involved with {} {:02x?}.",
+                    counter_party,
+                    stringify!(Channel),
+                    channel.get_id()
+                )));
+            }
+            match channel {
+                Channel::Offered(offered_channel) => {
+                    // remove rejected channel, since nothing has been confirmed on chain yet.
+                    self.store.upsert_channel(Channel::Cancelled(offered_channel), None)?;
+                },
+                Channel::Signed(mut signed_channel) => {
+                    crate::channel_updater::on_reject(&mut signed_channel)?;
 
-        self.store
-            .upsert_channel(Channel::Signed(signed_channel), None)?;
+                    self.store
+                        .upsert_channel(Channel::Signed(signed_channel), None)?;
+                },
+                channel => {
+                    return Err(Error::InvalidState(
+                        format!("Not in a state adequate to receive a reject message. {:?}", channel),
+                    ))
+                }
+            }
+        } else {
+            warn!("Couldn't find rejected dlc channel with id: {}", reject.channel_id.to_hex());
+        }
+
         Ok(())
     }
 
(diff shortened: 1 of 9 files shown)


dlc-manager/src/manager.rs
```diff
@@ -1941,14 +1953,39 @@ where
...
+            if channel.get_counter_party_id() != *counter_party {
+                return Err(Error::InvalidParameters(format!(
+                    "Peer {:02x?} is not involved with {} {:02x?}.",
+                    counter_party,
+                    stringify!(Channel),
+                    channel.get_id()
+                )));
+            }
+            match channel {
+                Channel::Offered(offered_channel) => {
```
Tibo-lg, line 1969: I think we'd need to unlock the utxos here, but it's a bit a big change so let me know if you want to do that here or just add an issue and someone can do it later.

holzeis: Good catch! I will have a look.

holzeis:
@Tibo-lg I've addressed your review remark in 19b0e87cc39386f2e0fd041002d904e4cdd64ea0

Additionally, I noticed that in my original commit I haven't updated the offered contract state to rejected. I've fixed that here ec83983e1fd5fd8698b211321a254ab4d0461efc.

Please have another look.

dlc-manager/src/manager.rs
holzeis: Addressed in 19b0e87cc39386f2e0fd041002d904e4cdd64ea0

dlc-manager/src/manager.rs
```diff
@@ -924,6 +926,31 @@ where
         Ok(msg)
     }
 
+    /// Reject a channel that was offered. Returns the [`dlc_messages::channel::Reject`]
+    /// message to be sent as well as the public key of the offering node.
+    pub fn reject_channel(&self, channel_id: &ChannelId) -> Result<(Reject, PublicKey), Error> {
+        let offered_channel = get_channel_in_state!(self, channel_id, Offered, None as Option<PublicKey>)?;
+        let offered_contract = get_contract_in_state!(self, &offered_channel.offered_contract_id, Offered, None as Option<PublicKey>)?;
+
+        if offered_channel.is_offer_party {
```
Tibo-lg, line 935: So the offer party can also cancel the offer by calling reject? I hadn't realized this, if that's intended it might be good to reflect this in the method name and documentation.

dlc-manager/src/manager.rs
```diff
@@ -1986,14 +2013,39 @@ where
...
+                    "Peer {:02x?} is not involved with {} {:02x?}.",
+                    counter_party,
+                    stringify!(Channel),
+                    channel.get_id()
+                )));
+            }
+            match channel {
+                Channel::Offered(offered_channel) => {
+                    // remove rejected channel, since nothing has been confirmed on chain yet.
+                    self.store.upsert_channel(Channel::Cancelled(offered_channel), None)?;
```
Tibo-lg, line 2031: Shouldn't you also change the state of the contract?

dlc-manager/tests/channel_execution_tests.rs
```diff
@@ -101,6 +101,7 @@ enum TestPath {
     RenewReject,
     RenewRace,
     RenewEstablishedClose,
+    CancelOffer,
```
Tibo-lg, line 104: If both offer and accept can cancel it might be good to test both cases.

Tibo-lg (commented):
Thanks for adding the `unreserve_utxo`! Just a couple more things

dlc-manager/src/manager.rs
holzeis: Nope, that was not intentional 🙈, although probably a good feature to add as well later.

I fixed it in 645b9bfb95b722de06251d3077a1c1b997eaa7f6

dlc-manager/src/manager.rs
holzeis: Yes, I should :)

fd34a33a57a829a5143a2bb2ca84a3e90dca23f0

dlc-manager/tests/channel_execution_tests.rs
holzeis: see comment above, that was not intentional.

holzeis:
@Tibo-lg Thanks for your review. I've addressed your remarks in 645b9bfb95b722de06251d3077a1c1b997eaa7f6 and fd34a33a57a829a5143a2bb2ca84a3e90dca23f0.

Please have another look.

dlc-manager/src/manager.rs
```diff
@@ -1986,14 +2001,50 @@ where
...
+
+                    // remove rejected channel, since nothing has been confirmed on chain yet.
+                    self.store.upsert_channel(Channel::Cancelled(offered_channel), Some(Contract::Rejected(offered_contract)))?;
+                },
+                Channel::Signed(mut signed_channel) => {
+                    crate::channel_updater::on_reject(&mut signed_channel)?;
 
-        crate::channel_updater::on_reject(&mut signed_channel)?;
+                    self.store
+                        .upsert_channel(Channel::Signed(signed_channel), None)?;
```
Tibo-lg, line 2036: I think it was like that before, but I now realized that in the case of rejecting in `RenewOffer` the offered contract state should also be rejected.

dlc-manager/src/manager.rs
```diff
@@ -924,6 +926,19 @@ where
         Ok(msg)
     }
 
+    /// Reject a channel that was offered. Returns the [`dlc_messages::channel::Reject`]
+    /// message to be sent as well as the public key of the offering node.
+    pub fn reject_channel(&self, channel_id: &ChannelId) -> Result<(Reject, PublicKey), Error> {
```
Tibo-lg, line 931: If this should be called only by the accept party, it feels like it would be good to have a check for that.

Tibo-lg (commented):
Sorry some more small things

dlc-manager/src/manager.rs
holzeis: 300cbf1dd884f0fa8dcbfb6356604699880f06e0

dlc-manager/src/manager.rs
holzeis: 300cbf1dd884f0fa8dcbfb6356604699880f06e0

holzeis:
@Tibo-lg I've addressed your additional remarks in 300cbf1dd884f0fa8dcbfb6356604699880f06e0

Please have another look

Tibo-lg (approved):
Thank you!

holzeis:
> Thank you!

oh you've merged it already 😅 . I would have cleaned up the fixups, before merging.. :)

Tibo-lg:
> oh you've merged it already 😅 . I would have cleaned up the fixups, before merging.. :)

Somehow I thought github was clever enough to squash them on merge, I guess I keep expecting too much from them...

yancyribbens:
> Somehow I thought github was clever enough to squash them on merge, I guess I keep expecting too much from them...

Can't tell if this is sarcasm but there is a way to squash and merge.  just click the green drop down arrow next to "rebase and merge" and choose "squash and merge" instead.

Tibo-lg:
Yeah squashing is not really what I want to, I'd like it to do an autosquash, but I guess that's too much to ask :)

yancyribbens:
What's the difference between squashing and auto squashing?

yancyribbens:
hmm, cool.  Thanks!

PR #203, opened 2024-02-06
Remove global context use from dlc-manager

Tibo-lg:
I noticed that #188 broke the build of `dlc-manager` (when building from the folder, not the workspace) because of using the secp global-context while not importing it as a feature. As I'm not super keen on adding extra features I removed the need for it with some refactoring.

 @benthecarman can you have a look and make sure it still works for you?


Tibo-lg:
This is more tricky than I thought, need some time to fix it properly.

Tibo-lg:
Should be good like that. Note that the compilation only fails after updating the dependencies (#197), really unsure why.

benthecarman (approved):
Looks good to me

PR #212, opened 2024-04-03
Change Postgres version to 12.2 for sample

bennyhodl:
Fixes issue when starting postgres db
```
2024-04-03 15:12:44.324 UTC [1] FATAL:  database files are incompatible with server
2024-04-03 15:12:44.324 UTC [1] DETAIL:  The data directory was initialized by PostgreSQL version 12, which is not compatible with this version 16.2 (Debian 16.2-1.pgdg120+2).
```

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -1,4 +1,9 @@
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+      - 'testci/**'
+  pull_request:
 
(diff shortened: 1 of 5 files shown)


Tibo-lg:
This is most likely because you have an volume leftover from a run with the old version of the docker compose file. Try running a `docker compose --profile oracle down -v` and then start again and it should work (assuming you don't care about the data on the db volume).

Issue #221, opened 2024-05-30
Upgrade `rust-bitcoin` to `0.32.0`


bennyhodl:
I am consuming `rust-dlc` with `bdk`. Bdk has an alpha version of 1.0 which is going to use `rust-bitcoin = "0.32"`

Taking a stab at this per crate.

Track BDK 0.32 process: https://github.com/orgs/bitcoindevkit/projects/14?pane=issue&itemId=61724768

PR #226, opened 2024-06-05
Remove reliance on `secp256k1_zkp`

bennyhodl:
`secp256k1_zkp` complicates updating `bitcoin` to new minor version. Issues that I faced with #223. This is a better starting point for #221 & #223 will be closed if this is merged.

Using `secp256k1_zkp` does not make much sense when it is exported from `bitcoin`. Changes all references except the `EcdsaAdaptorSignature`. Would like options to remove `secp256k1_zkp` completely.

- **deps: dlc to use secp exported from bitcoin**
- **deps: dlc-manager to use secp exported from bitcoin**
- **deps: dlc-messages to use exported secp from bitcoin**
- **deps: dlc-sled-storage-provider to use exported secp from bitcoin**
- **deps: dlc-trie to use exported secp from bitcoin**
- **deps: mocks to use exported secp from bitcoin**
- **deps: p2pd-oracle-client to use exported secp from bitcoin**
- **deps: simple-wallet to use exported secp from bitcoin**


luckysori:
@bennyhodl: Perhaps I'm missing the point, but why does it matter where you import the types from?

Whether you import transitively through `bitcoin` or transitively through `secp256k1_zkp`, you're still getting the same types.

> Would like options to remove secp256k1_zkp completely

This seems impossible since we need to use ECDSA adaptor signatures to implement the DLC channel protocol.

bennyhodl:
> Whether you import transitively through bitcoin or transitively through secp256k1_zkp, you're still getting the same types.

Inline with the ₿itcoin crate is best imo. For example, we cannot do #221 becuase `secp256k1_zkp` does not have a version compatible with the secp crate exported in `bitcoin`.

> This seems impossible since we need to use ECDSA adaptor signatures to implement the DLC protocol.

It should be opened in broader discussion to implement adaptor signatures to `bitcoin`. They are both ~the same~ library.

luckysori:
> For example, we cannot do #221 becuase `secp256k1_zkp` does not have a version compatible with the secp crate exported in `bitcoin`.

Wouldn't it be simpler to ask the `secp256k1_zkp` maintainers to update the `secp256k1` dependency?

I think what you are proposing here may not even work because, once you bump `rust-bitcoin`, you will very likely end up needing to use a `secp256k1:0.28.0` type returned by `secp256k1_zkp` with an API that now expects `secp256k1:0.29.0` (or vice versa). I assume you will run into `mismatched types: perhaps two different versions of crate secp256k1 are being used?`

> It should be opened in broader discussion to implement adaptor signatures to `bitcoin`. They are both ~the same~ library.

Yeah, if ECDSA adaptor signatures are added to `secp256k1` then things would be easier. But I don't think this is happening any time soon.

bennyhodl:
Moved to draft until `zkp` has adaptor signatures

PR #228, opened 2024-06-16
chore: use correct feature name for serde

bennyhodl:
Replaces the feature `serde` to use the correct feature `use-serde`

Adds `use-serde` to `OracleAttestation`


Tibo-lg (approved):
Thanks!

PR #230, opened 2024-07-09
Update bitcoin node version

bennyhodl:
Adds a script `gen-sample-contract.sh` that creates a contract an hour in the future instead of manually editing.

Updates the bitcoin-core image to the latest. In doing this, the wallets have to be legacy for the RPC client to work. The RPC client will have to update to use descriptors in the `derive_contract_signer` method.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3.4"
 services:
   ### BITCOIND
   bitcoind:
-    image: ruimarinho/bitcoin-core:0.20.0
+    image: polarlightning/bitcoind:27.0
     container_name: bitcoin-node
     command: |
       -conf=/config/default.conf
--- a/sample/Readme.md
+++ b/sample/Readme.md
@@ -8,6 +8,8 @@ Example configurations and contract input are available in the [examples](./exam
 In order to use the [example contract](./examples/contracts/numerical_contract_input.json) you will need to update the `event_id`.
 Replace the part after `btcusd` with a UNIX timestamp in the future.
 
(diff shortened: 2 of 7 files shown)


docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.4"
 services:
   ### BITCOIND
   bitcoind:
-    image: ruimarinho/bitcoin-core:0.20.0
+    image: kylemanna/bitcoind:latest
```
Tibo-lg, line 5: I don't like to use `latest` because things then break unexpectedly.

sample/Readme.md
```diff
@@ -8,6 +8,8 @@ Example configurations and contract input are available in the [examples](./exam
 In order to use the [example contract](./examples/contracts/numerical_contract_input.json) you will need to update the `event_id`.
 Replace the part after `btcusd` with a UNIX timestamp in the future.
 
+Use the [helper script](../scripts/gen-sample-offer.sh) to create update the example contract 1 hour in the future. This replaces the `event_id` after `btcusd` with a UNIX timestamp 1 hour in the futre. Manually update the value for a longer time period.
```
Tibo-lg, line 11:
```suggestion
Use the [helper script](../scripts/gen-sample-offer.sh) to create the example contract 1 hour in the future. This creates an `event_id` with a UNIX timestamp 1 hour in the future. Manually update the value for a longer time period.
```

Tibo-lg (changes requested):
Looks great apart from the small comments. Also CI doesn't pass can you fix it?

Tibo-lg:
If you rebase on master the clippy errors should be gone.

Tibo-lg (approved):

PR #231, opened 2024-07-10
Fix f64 serialization

### LINKED ISSUE #216: Failed to roundtrip `dlc_messages::ser_impls::{write, read}_f64`
In our fork of `rust-dlc`, I just tried to use `write_f64` and `read_f64` in a roundtrip test and the test failed:

```
---- message_handler::tests::funding_fee_event_roundtrip stdout ----
thread 'message_handler::tests::funding_fee_event_roundtrip' panicked at crates/xxi-node/src/message_handler.rs:1110:9:
assertion `left == right` failed
  left: FundingFeeEvent { due_date: 100000, funding_rate: 0.0003, price: 60000.0, funding_fee: 100 }
 right: FundingFeeEvent { due_date: 100000, funding_rate: 0.0002899169921875, price: 60000.0, funding_fee: 100 }
```

In this example, both `funding_rate` and `price` are `f64`s.

These read and write functions ~aren't used or tested, so that's probably why this wasn't caught before~ are used by certain macros.

---

I can provide a minimal repro later, I just wanted to report the issue before I forget.

Tibo-lg:
Closes #216

### DIFF
--- a/dlc-messages/src/ser_impls.rs
+++ b/dlc-messages/src/ser_impls.rs
@@ -174,25 +174,22 @@ pub fn write_f64<W: lightning::util::ser::Writer>(
     input: f64,
     writer: &mut W,
 ) -> Result<(), ::lightning::io::Error> {
-    let sign = input >= 0.0;
-    sign.write(writer)?;
-    let input_abs = f64::abs(input);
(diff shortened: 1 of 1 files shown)


luckysori (approved):
Love it, thank you!

PR #234, opened 2024-09-09
Update to `bitcoin:0.32`

bennyhodl:
All peer dependencies are now able to update to `bitcoin` 32.

Most changes are renaming of methods and different export paths.

Notable changes:
* `lightning` no longer exports `std::io::Cursor` and uses an emulation through `lightning::io::Cursor`. PR [`lightning/src/io/mod.rs`](https://github.com/lightningdevkit/rust-lightning/commit/176d2ad599d04f31c1acdd7cbee8e2510721408d)
* New methods to the lightning peer manager `peer_connected` and `peer_disconnected`. I believe they don't do anything since there is no specific node feature flagging for DLC's. Which could be an interesting feature to add to assert peers are able to parse DLC mesages.
* When parsing network related information for writing addresses in the messenger, we can no longer parse the network since it is not public on the address. Since non-mainnet addresses are parsed the same we can only check if the network is mainnet or a testnet flavor. Not sure if we can pass a network here.
* `bitcoin` 32 introduces using `Amount` over `u64` in all transaction related structs. I opted to keep this crate as `u64` but I think should introduce using `Amount` over `u64` in a future PR.

### DIFF
--- a/dlc-manager/src/contract_updater.rs
+++ b/dlc-manager/src/contract_updater.rs
@@ -2,7 +2,7 @@
 
 use std::ops::Deref;
 
-use bitcoin::psbt::PartiallySignedTransaction;
+use bitcoin::psbt::Psbt;
 use bitcoin::{consensus::Decodable, Script, Transaction, Witness};
 use dlc::{DlcTransactions, PartyParams};
 use dlc_messages::FundingInput;
@@ -120,7 +120,7 @@ where
         &accept_params,
         &funding_inputs,
         &signer.get_secret_key()?,
-        fund_output_value,
+        fund_output_value.to_sat(),
         None,
         &dlc_transactions,
     )?;
@@ -282,7 +282,7 @@ where
         &accept_msg.funding_inputs,
         &accept_msg.refund_signature,
         &cet_adaptor_signatures,
-        fund_output_value,
+        fund_output_value.to_sat(),
         wallet,
         &signer,
         None,
@@ -296,10 +296,7 @@ where
     Ok((signed_contract, signed_msg))
 }
 
-fn populate_psbt(
-    psbt: &mut PartiallySignedTransaction,
-    all_funding_inputs: &[&FundingInput],
-) -> Result<(), Error> {
+fn populate_psbt(psbt: &mut Psbt, all_funding_inputs: &[&FundingInput]) -> Result<(), Error> {
     // add witness utxo to fund_psbt for all inputs
     for (input_index, x) in all_funding_inputs.iter().enumerate() {
         let tx = Transaction::consensus_decode(&mut x.prev_tx.as_slice()).map_err(|_| {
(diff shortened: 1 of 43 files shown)


bennyhodl:
`create_cet_adaptor_sig_is_valid` is failing. Couldn't track down why.

Tibo-lg:
That's nice thanks! I couldn't see right away what could be the issue with the unit test that is failing, but it might be good to try to get the integration tests running to see if it's really just that unit test or if there is an actual issue (not sure what is the issue with them, seems something to do with coin selection).

simple-wallet/src/lib.rs
```diff
@@ -249,14 +245,20 @@ where
...
-                satisfaction_weight: 107,
+            .map(|x| {
+                WeightedUtxo {
+                    utxo: BdkUtxo::Local(LocalOutput {
+                        outpoint: x.outpoint,
+                        txout: x.tx_out.clone(),
+                        keychain: KeychainKind::External,
+                        is_spent: false,
+                        // @Tibo how do we want to handle this? Store the localoutput in storage?
+                        confirmation_time: ConfirmationTime::unconfirmed(1),
```
bennyhodl, line 256: @Tibo-lg coin selection seems to be the issue for integration tests. Doesn't seem to lead to the unit test though.

Integration tests are very flakey so haven't gotten them to fully pass.

How should we handle the derivation index and conf time? Not sure if it should be included with the `Utxo` type.

Tibo-lg:
I think I've fixed all issues. I wanted to push to your branch but was somehow not able not sure why. Can you cherry pick this commit? https://github.com/p2pderivatives/rust-dlc/commit/be327ce677170215f6b484d2238f099be4575629
Then have a look to make sure the changes make sense to you. I'll have a last look tomorrow and if everything is good will merge.

bennyhodl:
tACK

rebased and pushed with the changes.

Looks like the issue was the hash type macro?

bennyhodl:
Is this good to merge and release?

Tibo-lg:
> Is this good to merge and release?

No CI is not passing, it seems there is some PSBT issue. I don't have time to look at it right now, I'll try to next week, but if you solve it before I'm happy to make a release
https://github.com/p2pderivatives/rust-dlc/actions/runs/10863236537/job/30147194064?pr=234#step:7:13

dlc-manager/src/contract_updater.rs
```diff
@@ -655,7 +656,14 @@ where
         channel_id,
     };
 
-    Ok((signed_contract, fund_psbt.extract_tx()))
+    let transaction = fund_psbt.extract_tx().map_err(|e| {
+        Error::InvalidState(format!(
+            "Could not extract transaction from funding psbt. {}",
+            e
+        ))
+    })?;
```
Tibo-lg, line 664: This should fix the error.
```suggestion
        let transaction = fund_psbt.extract_tx_unchecked_fee_rate();
```

Tibo-lg (approved):

PR #235, opened 2024-10-04
feat: async support for dlc-manager

bennyhodl:
builds on top of https://github.com/p2pderivatives/rust-dlc/pull/234

Adds a feature flag of `async`. There are functions for less reliance on oracles for creating offers and closing contracts. In my implementation of `dlcdevkit` the consumers are typically going to use async runtimes with gRPC servers, REST servers, tauri, etc. Manually closing contracts and tracking for attestations created code and solutions that I was not happy with for something I believe should be handled internally in the library.

I think that `dlc-manager` should manage contracts agnostic to runtimes. I.e. not having to spawn a blocking thread for periodic checks.

This makes the manual close for situations in which an attestation is produced before the maturity. Not necessarily for async runtimes that are not running periodic checks for confirmed contracts.


Tibo-lg:
Having async interfaces would definitely be nice but that's something I'd like to give some thinking rather than just duplicating code tbh. I know bdk and ldk are using some macros to make this better and I'd probably want to check that first before considering which approach to take.

bennyhodl:
Understood, looking at the [`bdk-macros`](https://github.com/bitcoindevkit/bdk/blob/v0.29.0/macros/src/lib.rs) crate it just supports for wasm targets. Would have to fork the code for the `async` feature. We could have the maybe async oracle trait and then call `[maybe_await]` for the calls for attestations and announcements. 

Will have to change all of the iterators to for loops to keep the calls to the Oracle traits the same.

Would this be worthwhile to try?

PR #237, opened 2024-10-14
Check balances script

alfonzii:
Added check_balances.sh script to print out wallets balances for further experimenting


Tibo-lg (approved):
I guess it's fine but tbh I'm not sure we need a script for every single bitcoin-cli command. Sometimes it's also ok to just directly invoke the command from the terminal :)

Issue #239, opened 2024-10-28
spec: Oracle Attestation


bennyhodl:
The oracle attestation should include the `event_id` and `nb_signatures`

https://github.com/discreetlogcontracts/dlcspecs/blob/master/Messaging.md#the-oracle_attestation-type

Tibo-lg:
Yeah the master branch hasn't been updated to the latest specs. There is a branch that has the necessary changes here: https://github.com/p2pderivatives/rust-dlc/tree/feature/new-oracle-format-regular (but it's quite behind).

Issue #181, opened 2023-12-01
Add `validate` to `OracleAttestation`


benthecarman:
I just spent too long debugging an issue where I put all the possible outcomes in an `OracleAttestation` instead of the actual outcome. 

The validate could just validate that sigs.len() == outcomes.len() but it could also verify the sigs. Further it could also check it used the correct nonces.

PR #242, opened 2024-10-29
OracleAttestation validation and follow spec

bennyhodl:
Adds validation to OracleAttestation and adds `event_id` to the messaging to follow spec.

Closes https://github.com/p2pderivatives/rust-dlc/issues/239
Closes https://github.com/p2pderivatives/rust-dlc/issues/181


bennyhodl:
I think the failing tests in sled are because the sled files do not have the event_id in the attestation?

Tibo-lg:
> I think the failing tests in sled are because the sled files do not have the event_id in the attestation?

Yeah you should be able to solve this by running the `generate_serialized_contract_files.sh` script.

bennyhodl:
I added attestation validation in line with the [spec](https://github.com/discreetlogcontracts/dlcspecs/blob/master/Oracle-Validation.md#requirements-2)

Add two test of a valid attestation and an invalid attestation where signature does not match the nonce.

Tibo-lg (approved):
Thanks!

PR #243, opened 2024-10-31
Fix OracleEvent.validate() for signed numeric events

rorp:
For some reason `OracleEvent.validate()` doesn't count a nonce for the sign. This PR fixes this.


Tibo-lg (commented):
Thanks! Do you think that you could add a test that fails without the change and passes with it?

Tibo-lg (approved):
Thanks!

bennyhodl:
Is this going to merge and release?

https://github.com/bennyhodl/kormir/issues/14

PR #244, opened 2024-11-03
feat: automated releases w/ release-plz

bennyhodl:
It would be nice to have automated releases on merged PRs. I have used release-plz in another project and it is great.

- uses github actions
- generates changelogs
- creates tags & releases
- follow semver to automatically update version
- publishes to crates

All that is needed is to add the `CARGO_REGISTRY_TOKEN` and `RELEASE_PLZ_TOKEN` (which is a GH token) added to the repository secrets. @Tibo-lg can add them or if you would like me to manage releases, you can add me as a repo actions owner and a package owner on crates.

Useful links:
[release-plz docs](https://release-plz.ieni.dev/)
[release-plz github actions docs](https://release-plz.ieni.dev/docs/github/quickstart)
[release-plz configuration](https://release-plz.ieni.dev/docs/config)

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -1,0 +1,52 @@
+name: Release-plz
+
+permissions:
+  pull-requests: write
+  contents: write
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  release-plz-release:
+    name: Release-plz release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          token: ${{ secrets.RELEASE_PLZ_TOKEN }}
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run release-plz
+        uses: MarcoIeni/release-plz-action@v0.5
+        with:
+          command: release
+        env:
+          GITHUB_TOKEN: ${{ secrets.RELEASE_PLZ_TOKEN }}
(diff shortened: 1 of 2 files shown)


Tibo-lg:
Thanks but I don't especially like this. While convenient I would need to trust what it is doing to the code and the policy is to minimize dependencies in general to avoid security issues.

bennyhodl:
`release-plz` under the hood uses `std::process::Command` for git history and publishing to cargo. [Even the rust-foundation uses it](https://github.com/rust-lang/libm/blob/master/.github/workflows/publish.yml)

For the past few months I have had to ask for releases and have had to use patches exclusively for `ddk`. This is not sustainable for downstream projects. I would rather have a conversation about it than immediately close. 

I think this is a net benefit for the project as you do not have to manually do releases, tags, change logs, and publish given your limited time.

bonomat:
Any reason why you (@bennyhodl ) are not depending on a specific git commit until there is a release? 
This worked great for us at 10101 in the past.

bennyhodl:
> Any reason why you (@bennyhodl ) are not depending on a specific git commit until there is a release? 
> 
> This worked great for us at 10101 in the past. 

Because ddk is a library I cannot publish to crates.io with a fit link

Tibo-lg:
In general I try to follow the processes of other crates in the rust-bitcoin ecosystem. AFAICT none of them are using this, and giving the ability for a third party tool to publish on creates.io is not something that feels secure to me. If you want to help with releases, feel free to make release PRs so that I can check them and merge them. Doing the release on `crates.io` doesn't take much time, it's more the preparation that does.

PR #245, opened 2024-11-03
check: fix benches with deps upgrade

bennyhodl:
benchmarks were not updated when upgraded to bitcoin 32


Tibo-lg (approved):
Thanks! Should probably build them as part of CI at some point.

PR #246, opened 2024-11-11
Minor improvements to sample Readme

optout21:
Some typos and minor explanations

### DIFF
--- a/sample/Readme.md
+++ b/sample/Readme.md
@@ -8,15 +8,15 @@ Example configurations and contract input are available in the [examples](./exam
 In order to use the [example contract](./examples/contracts/numerical_contract_input.json) you will need to update the `event_id`.
 Replace the part after `btcusd` with a UNIX timestamp in the future.
 
-Use the [helper script](../scripts/gen-sample-offer.sh) to create the example contract 1 hour in the future. This creates an `event_id` with a UNIX timestamp 1 hour in the future. Manually update the value for a longer time period.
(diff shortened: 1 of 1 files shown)


sample/Readme.md
```diff
@@ -17,7 +17,7 @@ To give a quick try to this sample, run the following set of commands (assuming
 
 ```bash
 docker-compose --profile oracle up -d
-docker compose exec bitcoind /scripts/create_wallets.sh
+docker-compose exec bitcoind /scripts/create_wallets.sh
```
Tibo-lg, line 20: I think `docker compose` is the correct syntax for new docker versions. IIRC `docker-compose` was for older version. So maybe it would be better to fix the line above.

Tibo-lg (changes requested):
Small nit with `docker` otherwise nice thanks!

sample/Readme.md
optout21: I'm not sure which one is proper, but on my default docker install on recent Ubuntu, there is no `compose` command.
`Docker version 24.0.7, build 24.0.7-0ubuntu4.1`
Maybe `docker compose` with a note that `docker-compose` is an (older) alternative?

Tibo-lg: Looking into it a bit more, it seems like `docker-compose` was for compose v1 which is apparently deprecated. I'm fine with having a note for it though it you find it helpful.
https://docs.docker.com/compose/releases/migrate/

optout21: Note that `docker-compose` is used and mentioned in 2 script files and 2 doc files as well, I'm not touching those here.
This PR updated as suggested,

Tibo-lg (approved):

PR #252, opened 2025-01-22
fix: prealloc with making CETs

jharveyb:
Most other vectors near this point in the library are pre-allocated, but this one wasn't. I observed a small (0.3%) perf increase for the existing benchmarks, when adapted to run under `iai-callgrind` instead of Criterion.


Tibo-lg (approved):
Thanks!

PR #255, opened 2025-03-16
feat: close_dlc msg and cooperative close logic

matthewjablack:
## What

Add `close_dlc` msg type, and cooperative close logic

## Why

It allows Alice or Bob to pass a message with the intention of cooperatively closing the DLC, allowing for early exit from a position.

The initiator provides an input, so they can cancel if the counterparty doesn't respond right away.

Yes this creates a free option problem, but the expectation is that the counterparty is a trusted party (for example a market maker). 

This message type was discussed in DLC Specs: https://github.com/discreetlogcontracts/dlcspecs/issues/161

And has been implemented in the Atomic Finance typescript libraries: https://github.com/AtomicFinance/bitcoin-abstraction-layer/pull/86 and https://github.com/AtomicFinance/node-dlc/pull/122


dlc-manager/src/manager.rs
```diff
@@ -470,6 +474,47 @@ where
...
+                    close_msg.offer_payout.to_sat() as i64 -
+                    signed_contract.accepted_contract.offered_contract.offer_params.collateral.to_sat() as i64
+                } else {
+                    close_msg.accept_payout.to_sat() as i64 -
+                    signed_contract.accepted_contract.accept_params.collateral.to_sat() as i64
+                }
+            ),
+        };
+
+        self.store.update_contract(&Contract::Closed(closed_contract))?;
```
Tibo-lg: Would be better to use `PreClosed` here. And so set `signed_cet` to the closing tx.

dlc-manager/src/contract_updater.rs
```diff
@@ -754,6 +755,129 @@ where
...
+        fund_input_serial_id: offered_contract.fund_output_serial_id,
+        funding_inputs: accepted_contract.funding_inputs.clone(),
+        funding_signatures: signed_contract.funding_signatures.clone(),
+    };
+
+    Ok((close_message, close_tx))
+}
+
+/// Verifies and completes a cooperative close transaction using the counter party's signature.
+pub fn verify_and_complete_cooperative_close<C: Signing, SP: Deref>(
```
Tibo-lg: nit: I don't think it's verifying anything here so maybe you can drop the `verify_`

dlc-manager/src/contract/mod.rs
```diff
@@ -43,6 +43,8 @@ pub enum Contract {
     Closed(ClosedContract),
     /// A contract whose refund transaction was broadcast.
     Refunded(signed_contract::SignedContract),
+    /// A contract that was closed cooperatively.
+    CooperativeClose(CooperativeCloseContract),
```
Tibo-lg: What is the point of that state? I feel the best would be to just not updated the state when proposing, and then update if/when the closing transaction is seen on-chain. Right now if you propose a cooperative close but it somehow doesn't end up being closed you will be in an incorrect state.

dlc-manager/src/manager.rs
```diff
@@ -470,6 +474,47 @@ where
         Ok(())
     }
 
+    fn on_close_message(&self, close_msg: &CloseDlc, counter_party: &PublicKey) -> Result<(), Error> {
+        let signed_contract = get_contract_in_state!(
+            self,
+            &close_msg.contract_id,
+            Signed,
```
Tibo-lg: Should be `Confirmed`

Tibo-lg (changes requested):
It's a good feature, though:
* please add an integration test so we can be confident it's working, I don't think it does right now
* The chain monitor is used for channels, I don't think what you implemented here will work
* I think the best way to go to close the contract on the party offering the cooperative close would be to stay in the `Confirmed` state, but then maybe have a `close_tx: Option<Transaction>` field that you check within `check_confirmed_contract` so that you can mark the contract as `PreClosed/Closed`. It could also be a list if we assume that there could be multiple cooperative close offers (in case some are not accepted/processed).

dlc-manager/src/manager.rs
matthewjablack: Good catch, fixed in [22c22c6](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/22c22c62d2f653ed68840f14d26ca958d783ac62)

dlc-manager/src/contract/mod.rs
matthewjablack: Agreed, removed this state in [22c22c6](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/22c22c62d2f653ed68840f14d26ca958d783ac62)

dlc-manager/src/contract_updater.rs
matthewjablack: Good point, fixed in [22c22c6](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/22c22c62d2f653ed68840f14d26ca958d783ac62)

dlc-manager/src/manager.rs
matthewjablack: Good catch, fixed in [22c22c6](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/22c22c62d2f653ed68840f14d26ca958d783ac62)

matthewjablack:
Thanks for that review @Tibo-lg 

> please add an integration test so we can be confident it's working, I don't think it does right now

Added to `manager_execution_tests.rs`

> The chain monitor is used for channels, I don't think what you implemented here will work

Thanks for clarification, removed accordingly

> I think the best way to go to close the contract on the party offering the cooperative close would be to stay in the Confirmed state, but then maybe have a close_tx: Option<Transaction> field that you check within check_confirmed_contract so that you can mark the contract as PreClosed/Closed. It could also be a list if we assume that there could be multiple cooperative close offers (in case some are not accepted/processed).

Not sure what you had in mind here. Were you thinking of having `close_tx: Option<Transaction>` be added to an existing struct? Which one did you have in mind? 

I'm guessing you were thinking of adding it to `DlcTransactions`

dlc-manager/src/manager.rs
```diff
@@ -470,6 +474,65 @@ where
...
+                            .accepted_contract
+                            .accept_params
+                            .collateral
+                            .to_sat() as i64
+                },
+            ),
+        };
+
+        self.store
+            .update_contract(&Contract::Closed(closed_contract))?;
```
Tibo-lg: PreClosed

dlc-manager/tests/manager_execution_tests.rs
```diff
@@ -806,93 +848,194 @@ fn manager_execution_test(test_params: TestParams, path: TestPath, manual_close:
...
+
+                    // Bob receives and accepts the cooperative close
+                    bob_manager_send
+                        .lock()
+                        .unwrap()
+                        .accept_cooperative_close(&contract_id, &close_msg)
+                        .expect("Error accepting cooperative close");
+
+                    // Bob should now be in Closed state (he broadcast the transaction)
+                    assert_contract_state!(bob_manager_send, contract_id, Closed);
```
Tibo-lg: Should be PreClosed because the transaction is not confirmed.

dlc-manager/tests/manager_execution_tests.rs
```diff
@@ -806,93 +848,194 @@ fn manager_execution_test(test_params: TestParams, path: TestPath, manual_close:
...
+
+                    // Alice should still be in Confirmed state (she doesn't know about the close yet)
+                    assert_contract_state!(alice_manager_send, contract_id, Confirmed);
+
+                    // Mine a block to confirm the close transaction
+                    generate_blocks(1);
+
+                    // In a real scenario, Alice would detect the close transaction and call on_counterparty_close
+                    // For the test, we'll verify the cooperative close functionality worked correctly
```
Tibo-lg: We definitely want to detect the close transaction, and we want to test it.

dlc-manager/src/manager.rs
```diff
@@ -470,6 +474,65 @@ where
         Ok(())
     }
 
+    fn on_close_message(
```
Tibo-lg, line 477: I'm not sure we should have that to be honest. It feels like having only the `accept` method is better. This means I can just send a close message and have it accepted by default.

Tibo-lg (changes requested):

Tibo-lg:
> Not sure what you had in mind here. Were you thinking of having close_tx: Option<Transaction> be added to an existing struct? Which one did you have in mind?

> I'm guessing you were thinking of adding it to DlcTransactions

I don't have a strong opinion on where to put it, I guess `DlcTransactions` struct is fine.

dlc-manager/src/manager.rs
matthewjablack: Agreed, I've updated it to simply validate the close message

dlc-manager/tests/manager_execution_tests.rs
matthewjablack: Good catch, fixed in 3356604

dlc-manager/src/manager.rs
matthewjablack: Thanks, fixed in 3356604

dlc-manager/tests/manager_execution_tests.rs
matthewjablack: I've added `check_pending_close_transactions` function and included it in `periodic_check` https://github.com/p2pderivatives/rust-dlc/pull/255/files#diff-6901f71d060cd9a08d68e0e45ba8e40bcd714a041ddcdff6ea168eb4b35a1c4dR813-R863

Also updated tests https://github.com/p2pderivatives/rust-dlc/pull/255/files#diff-11a961d721d78a3c54a7b2a0b90b92b5a3768c0fbed6fef9c45014c0be8702e8R910-R936

matthewjablack:
Thanks @Tibo-lg 

I added `pending_close_txs` field to `DlcTransactions` to track cooperative close offers, `check_pending_close_transactions()` for chain. monitoring

Also fixed proper state transitions, with Confirmed → PreClosed → Closed

dlc-manager/src/contract_updater.rs
```diff
@@ -754,6 +758,127 @@ where
...
+        &offered_contract.offer_params,
+        offer_payout,
+        &accepted_contract.accept_params,
+        counter_payout,
+        fund_outpoint,
+        fund_output_value,
+    );
+
+    // Get our private key and sign the transaction
+    let fund_private_key =
```
bennyhodl: Instead of using `signer_provider.get_secret_key_for_pubkey()` you can use `signer_provider.derive_contract_signer(offered_contract.keys_id)`

Channels are the only implementation that uses this method. And you don't have to have a conditional for the correct funding pubkey. The funding pubkey comes from deriving a contract signer with the keys id and is stored in `OfferedContract`

bennyhodl (changes requested):

dlc-manager/src/contract_updater.rs
matthewjablack: Good catch, fixed in ec6cb89

dlc-messages/src/channel.rs
```diff
@@ -573,7 +573,6 @@ impl_dlc_writeable!(CollaborativeCloseOffer, {
     derive(serde::Serialize, serde::Deserialize),
     serde(rename_all = "camelCase")
 )]
-
```
Tibo-lg, line 576: Undo

dlc-messages/src/lib.rs
```diff
@@ -500,12 +501,57 @@ impl_dlc_writeable!(SignDlc, {
...
+    /// The payout amount for the offer party in satoshis.
+    pub offer_payout: Amount,
+    /// The payout amount for the accept party in satoshis.
+    pub accept_payout: Amount,
+    /// Serial id for the funding input.
+    pub fund_input_serial_id: u64,
+    /// The funding inputs to use.
+    pub funding_inputs: Vec<FundingInput>,
+    /// The funding signatures.
+    pub funding_signatures: FundingSignatures,
```
Tibo-lg, line 534: I think these fields are never used?

dlc-messages/src/lib.rs
```diff
@@ -500,12 +501,57 @@ impl_dlc_writeable!(SignDlc, {
...
+            serialize_with = "crate::serde_utils::serialize_hex",
+            deserialize_with = "crate::serde_utils::deserialize_hex_array"
+        )
+    )]
+    /// The id of the contract to close.
+    pub contract_id: [u8; 32],
+    /// The signature for the closing transaction.
+    pub close_signature: Signature,
+    /// The payout amount for the offer party in satoshis.
+    pub offer_payout: Amount,
```
Tibo-lg: You don't need to pass that it can be re-computed from `total_collateral`

dlc-manager/src/manager.rs
```diff
@@ -895,6 +992,86 @@ where
...
+
+        let counter_party = signed_contract
+            .accepted_contract
+            .offered_contract
+            .counter_party;
+
+        Ok((close_message, counter_party))
+    }
+
+    /// Accepts a cooperative close request by verifying the counter party's signature,
```
Tibo-lg: I don't think you are verifying the signature (not that you need to, since if it's incorrect the broadcast will fail)

dlc-manager/tests/manager_execution_tests.rs
```diff
@@ -806,93 +848,211 @@ fn manager_execution_test(test_params: TestParams, path: TestPath, manual_close:
...
+                    // Both parties should now move to Closed state after full confirmations
+                    periodic_check!(bob_manager_send, contract_id, Closed);
+                    periodic_check!(alice_manager_send, contract_id, Closed);
+
+                    // Verify both parties are now in Closed state
+                    assert_contract_state!(bob_manager_send, contract_id, Closed);
+                    assert_contract_state!(alice_manager_send, contract_id, Closed);
+
+                    // Verify the close transaction was properly broadcast and confirmed
+                    let _close_txid = {
```
Tibo-lg: Remove this block if you don't need this

dlc-manager/tests/manager_execution_tests.rs
```diff
@@ -806,93 +848,211 @@ fn manager_execution_test(test_params: TestParams, path: TestPath, manual_close:
...
-                    }
-                }
-                TestPath::Refund => {
-                    periodic_check!(first, contract_id, Confirmed);
+                        (bob_manager_send, alice_manager_send)
+                    };
 
-                    periodic_check!(second, contract_id, Confirmed);
+                    match path {
+                        TestPath::Close => {
```
Tibo-lg: I would prefer that you don't refactor the rest of the tests to keep the PR focused, feel free to do another refactor PR.

dlc-manager/src/manager.rs
```diff
@@ -801,9 +770,26 @@ where
...
+                (None, pnl)
+            } else {
+                // CET close - include the signed_cet
+                let pnl = contract
+                    .signed_contract
+                    .accepted_contract
+                    .compute_pnl(&contract.signed_cet)?;
+                (Some(contract.signed_cet.clone()), pnl)
+            };
+
```
Tibo-lg: Why do you differentiate here?

dlc-manager/src/manager.rs
```diff
@@ -827,6 +810,57 @@ where
         Ok(())
     }
 
+    /// Check for pending cooperative close transactions
+    fn check_pending_close_transactions(&self) -> Result<(), Error> {
```
Tibo-lg: This logic should be implemented within `check_confirmed_contracts`, unless there is a reason to have it separate? Right now that means having to go two times through the confirmed contracts.

dlc-manager/src/manager.rs
```diff
@@ -827,6 +810,57 @@ where
...
+                let confirmations = self
+                    .blockchain
+                    .get_transaction_confirmations(&pending_close_tx.compute_txid())?;
+
+                if confirmations >= NB_CONFIRMATIONS {
+                    // Found a fully confirmed pending close - move directly to Closed
+                    let pnl = contract.accepted_contract.compute_pnl(pending_close_tx)?;
+                    let closed_contract = ClosedContract {
+                        attestations: None, // Cooperative close has no attestations
+                        signed_cet: None,   // Cooperative close doesn't use a CET
```
Tibo-lg: Why set it to None, I think it's fine to save the closing transaction here, unless it breaks something?

dlc-manager/src/manager.rs
Tibo-lg: Is the idea that an application should implement a custom message handler to accept it?

Tibo-lg (changes requested):

dlc-messages/src/channel.rs
matthewjablack: Undoing causes clippy errors

```
Run cargo clippy --no-default-features --features no-std -p dlc-messages -- -D warnings
warning: /home/runner/work/rust-dlc/rust-dlc/p2pd-oracle-client/Cargo.toml: no edition set: defaulting to the 2015 edition while the latest is 2024
warning: /home/runner/work/rust-dlc/rust-dlc/dlc-trie/Cargo.toml: no edition set: defaulting to the 2015 edition while the latest is 2024
warning: /home/runner/work/rust-dlc/rust-dlc/dlc/Cargo.toml: no edition set: defaulting to the 2015 edition while the latest is 2024
warning: /home/runner/work/rust-dlc/rust-dlc/bitcoin-test-utils/Cargo.toml: no edition set: defaulting to the 2015 edition while the latest is 2024
warning: /home/runner/work/rust-dlc/rust-dlc/dlc-messages/Cargo.toml: no edition set: defaulting to the 2015 edition while the latest is 2024
 Downloading crates ...
  Downloaded lightning-types v0.1.0
  Downloaded bech32 v0.9.1
  Downloaded possiblyrandom v0.2.0
  Downloaded hashbrown v0.13.2
  Downloaded lightning-invoice v0.32.0
  Downloaded libm v0.2.15
  Downloaded lightning v0.0.125
   Compiling secp256k1-sys v0.10.1
   Compiling bitcoin v0.32.7
   Compiling libc v0.2.174
   Compiling libm v0.2.15
    Checking cfg-if v1.0.1
    Checking bech32 v0.9.1
    Checking hashbrown v0.13.2
    Checking getrandom v0.2.16
    Checking possiblyrandom v0.2.0
   Compiling secp256k1-zkp-sys v0.10.1
    Checking secp256k1 v0.29.1
    Checking secp256k1-zkp v0.11.0
    Checking lightning-types v0.1.0
    Checking miniscript v12.3.5
    Checking lightning-invoice v0.32.0
    Checking lightning v0.0.125
    Checking dlc v0.7.1 (/home/runner/work/rust-dlc/rust-dlc/dlc)
    Checking dlc-messages v0.7.1 (/home/runner/work/rust-dlc/rust-dlc/dlc-messages)
error: empty line after outer attribute
   --> dlc-messages/src/channel.rs:571:1
    |
571 | / #[cfg_attr(
572 | |     feature = "use-serde",
573 | |     derive(serde::Serialize, serde::Deserialize),
574 | |     serde(rename_all = "camelCase")
575 | | )]
576 | |
    | |_^
577 |   /// Message used to reject an received offer.
578 |   pub struct Reject {
    |   ----------------- the attribute applies to this struct
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#empty_line_after_outer_attr
    = note: `-D clippy::empty-line-after-outer-attr` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::empty_line_after_outer_attr)]`
    = help: if the empty line is unintentional, remove it

error: could not compile `dlc-messages` (lib) due to 1 previous error
Error: Process completed with exit code 101.
```

Tibo-lg: Oh ok my bad!

dlc-messages/src/lib.rs
matthewjablack: These fields are necessary to ensure the offeror of the close contract can avoid a free option problem. They could offer a close to the counterparty, and the counterparty does not respond. 

https://github.com/discreetlogcontracts/dlcspecs/issues/161#issuecomment-867819121

matthewjablack: I've added an additional optional argument for `create_collaborative_close_transaction` to include additional inputs in [a915378](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/a9153786317c2270b6376a2cb351c3509849d7a9)

dlc-messages/src/lib.rs
matthewjablack: Great point, updated in [5e06374](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/5e06374092a8345fe2da40504585860f746b3ba9)

dlc-manager/src/manager.rs
matthewjablack: Good catch, updated the comment in [9d84b29](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/9d84b29b372bc2c7486a4127e0c538bbe6294a27)

dlc-manager/tests/manager_execution_tests.rs
matthewjablack: Good point, removed in [36c1e23](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/36c1e2394d22d4387cfce220cd5b30d13784ad76)

dlc-manager/tests/manager_execution_tests.rs
matthewjablack: Undid test refactor in [36c1e23](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/36c1e2394d22d4387cfce220cd5b30d13784ad76)

dlc-manager/src/manager.rs
matthewjablack: Moved to `check_confirmed_contracts` in [9d84b29](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/9d84b29b372bc2c7486a4127e0c538bbe6294a27)

dlc-manager/src/manager.rs
matthewjablack: You're right, signed_cet can be the closing tx, which is must more elegant. Updated in [9d84b29](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/9d84b29b372bc2c7486a4127e0c538bbe6294a27)

dlc-manager/src/manager.rs
matthewjablack: Good point, no need to differentiate, fixed in [9d84b29](https://github.com/p2pderivatives/rust-dlc/pull/255/commits/9d84b29b372bc2c7486a4127e0c538bbe6294a27)

dlc-manager/src/manager.rs
matthewjablack: Yep exactly. When closing a contract, the acceptor should double check that the payout matches their expected value of the contract, or lines up with their own business logic for them being comfortable with closing the contract early.

dlc-messages/src/lib.rs
Tibo-lg: Right now the way you set those fields does not make sense, you are setting them to the funding inputs and signatures, which matches with the names but not what you're describing here. If you want to be able to double spend the closing transaction then the inputs and signatures (actually a single one should be enough) should be set to utxos from the wallet. Honestly I feel the best is either you fully implement this (having it optional if you prefer), or just remove it.

PR #256, opened 2025-07-09
feat: single funded contracts

bennyhodl:
This PR implements single-funded DLCs where only one party provides collateral and covers all transaction fees, while the other party contributes no inputs or fees.

## Problem

The current implementation requires both parties to split transaction fees (funding, CET, and refund), even when one party provides zero collateral. This prevents true single-funded DLCs where the acceptor contributes nothing.

## Solution

Fee calculation is now dynamic based on collateral contribution:

- **Zero collateral**: Party pays no fees, provides no inputs
- **Full collateral**: Party pays all fees (single-funded)  
- **Partial collateral**: Fees split between parties (existing dual-funded behavior)

## Additional Changes

**Contract Input Validation**: Added dust limit checks per DLC specifications - offer collateral and total collateral must be above dust limit when non-zero. [Follow spec for offer collateral](https://github.com/discreetlogcontracts/dlcspecs/blob/9cd9148938c616690c79d99ec6f330e213c246c5/Protocol.md?plain=1#L141)

**Code Quality**: Fixed clippy warnings throughout the codebase.

## Backward Compatibility

Existing dual-funded DLCs work unchanged. API changes are additive with no breaking changes to existing functionality.

### DIFF
--- a/dlc-manager/src/contract/contract_input.rs
+++ b/dlc-manager/src/contract/contract_input.rs
@@ -8,6 +8,8 @@ use secp256k1_zkp::XOnlyPublicKey;
 #[cfg(feature = "use-serde")]
 use serde::{Deserialize, Serialize};
 
+const DUST_LIMIT: Amount = Amount::from_sat(1000);
+
 /// Oracle information required for the initial creation of a contract.
 #[derive(Debug, Clone)]
 #[cfg_attr(
@@ -87,6 +89,12 @@ pub struct ContractInput {
 impl ContractInput {
     /// Validate the contract input parameters
     pub fn validate(&self) -> Result<(), Error> {
+        if self.offer_collateral < DUST_LIMIT {
+            return Err(Error::InvalidParameters(
+                "Offer collateral must be greater than dust limit.".to_string(),
+            ));
+        }
+
         if self.contract_infos.is_empty() {
             return Err(Error::InvalidParameters(
                 "Need at least one contract info".to_string(),
--- a/dlc-manager/src/utils.rs
+++ b/dlc-manager/src/utils.rs
@@ -2,7 +2,7 @@
 use std::ops::Deref;
 
 use bitcoin::{consensus::Encodable, Amount, Txid};
-use dlc::{PartyParams, TxInputInfo};
+use dlc::{util::get_common_fee, PartyParams, TxInputInfo};
 use dlc_messages::{
     oracle_msgs::{OracleAnnouncement, OracleAttestation},
     FundingInput,
@@ -92,6 +92,7 @@ pub(crate) fn compute_id(
 pub(crate) fn get_party_params<W: Deref, B: Deref, X: ContractSigner, C: Signing>(
     secp: &Secp256k1<C>,
     own_collateral: Amount,
+    total_collateral: Amount,
     fee_rate: u64,
     wallet: &W,
     signer: &X,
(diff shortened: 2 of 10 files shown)


dlc-manager/src/contract/contract_input.rs
```diff
@@ -87,6 +89,19 @@ pub struct ContractInput {
...
     /// Validate the contract input parameters
     pub fn validate(&self) -> Result<(), Error> {
+        if self.offer_collateral < DUST_LIMIT {
+            return Err(Error::InvalidParameters(
+                "Offer collateral must be greater than dust limit.".to_string(),
+            ));
+        }
+
+        let total_collateral = self.offer_collateral + self.accept_collateral;
+        if total_collateral < DUST_LIMIT {
```
Tibo-lg: if `offer_collateral >= DUST_LIMIT` then `offer_collateral + accept_collateral >= DUST_LIMIT` so this check is unnecessary

dlc-manager/src/utils.rs
```diff
@@ -153,6 +153,37 @@ where
     Ok((party_params, funding_inputs))
 }
 
+// If own_collateral is zero, appr_required_amount is zero.
+// If own_collateral is equal to total_collateral, appr_required_amount is the common fee.
```
Tibo-lg, line 157: It feels like it would be easier to just pass a bool like `pay_all_fee` to this method

dlc/src/lib.rs
```diff
@@ -293,11 +293,22 @@ impl PartyParams {
...
         &self,
+        total_collateral: Amount,
         fee_rate_per_vb: u64,
         extra_fee: Amount,
     ) -> Result<(TxOut, Amount, Amount), Error> {
         let mut inputs_weight: usize = 0;
 
+        // first check if a party does not need to fund the contract if so, then it is zero
+        if self.collateral == Amount::ZERO {
+            // We use a zero value output to indicate that the party does not need to fund the contract
```
Tibo-lg, line 304: It would be clearer to return an option so that it can be `None` in that case.

dlc/src/lib.rs
```diff
@@ -427,9 +449,9 @@ pub(crate) fn create_fund_transaction_with_fees(
     let total_collateral = checked_add!(offer_params.collateral, accept_params.collateral)?;
 
     let (offer_change_output, offer_fund_fee, offer_cet_fee) =
-        offer_params.get_change_output_and_fees(fee_rate_per_vb, extra_fee)?;
+        offer_params.get_change_output_and_fees(total_collateral, fee_rate_per_vb, extra_fee)?;
```
Tibo-lg, line 452: Or maybe just don't call this method when it's not needed?

Tibo-lg (changes requested):
It's a nice feature to have! I feel the implementation and usage could be made a bit clearer, and also please add an integration test to make sure it's working as expected.

dlc-manager/src/utils.rs
bennyhodl: I'm not sure, to flag `pay_all_fee` then we still need to do the same check if they are paying all or zero collateral

dlc/src/lib.rs
bennyhodl: I disagree, making it an Option<T> requires extra checks in the callers that I think is cleaner to set to just Amount::ZERO

dlc/src/lib.rs
bennyhodl: You still need to include the accept_change to `create_fund_transaction` even if it is dust which is then discarded

Tibo-lg:
> The current implementation requires both parties to split transaction fees (funding, CET, and refund)

Going through this PR again, I wonder if it would make more sense to just have a setting that enables setting a single party to pay the entire fee regardless of the collateral values. It would also make things more explicit.

